PR #11, opened 2024-12-09
Move the main plugin file

### LINKED ISSUE #7: acf.php -> secure-custom-fields.php
If you're using this plugin as an opportunity to replace ACF and also replace the original SCF in the wp.org repo, why wouldn't you just start with a fresh file name?

I mean, if the goal is to actually maintain this version of a fork, why wouldn't you want to remove all instances of ACF both in filenames and in content?

kraftbj:
- **Move main plugin file. Closes #7**

For all new installs, it'll simply use a `secure-custom-fields.php` file. For existing installs, it'll either activate the new file on the single instance if only activated on a single site (if multisite) or across the network if network activated.

If activating the new file works, it'll deactivate the old file. If it can't activate the new file for whatever reason and the old one continues to load, there is an include_once so those sites should not see a functionality change.


j-hoffmann:
Well, this does not really fix #7. There, @robertdevore mentioned "removing all instances of ACF". This change leaves a legacy file `acf.php` in the plugin. Removing the file would break installs upgrading from current or past versions directly to future versions without the file (skipping versions with this fix). 

The ill-conceived move, of hastily pushing the new plugin to the wp.org plugin repo (as of now it has 500+ installs), lead to the ironical situation of this file having to remain in the plugin like a memorial to the ACF roots of this plugin.

kraftbj:
@j-hoffmann @robertdevore Nod. I've updated the description to say this will close that issue as it is incomplete as a fix for y'all.

For better or worse, this is a fork. Removing all references in code to `acf` is not a current priority. In comments, et al, it'll happen. New files will have new file names and as we rename existing files, we'll likely have legacy file names since it is relatively minor to do so versus breaking sites relying on things that assume particular file names, classes, function names, etc. ClassicPress is full of `wp` references, which, I think, makes total sense given the context.

j-hoffmann:
> breaking sites relying on things that assume particular file names

The most common thing where plugins use another plugin's main filename, is when checking whether they're active (`is_plugin_active( 'secure-custom-fields/acf.php' )`). This change breaks backwards compatibility to exactly that.

kraftbj:
@j-hoffmann Nod. With this code at `secure-custom-fields/acf.php` is relatively new, I'd rather make this change while things are in beta than any later. There's not a history with `secure-custom-fields/acf.php` as an active plugin to the degree that is present with `advanced-custom-fields/acf.php`. The class names, etc are a bit different since it can be retained and, tbh, I generally would do a `class_exists` check over an `is_plugin_active` check.

PR #14, opened 2024-12-13
Release: 6.4.1-beta4

kraftbj:
Update all version numbers to beta4.

Intending to update the version on w.org ASAP now that the main file has been updated to reduce anyone potentially impacted by the transition. See #11.


kraftbj:
* Confirmed built version's JS works without error.
* Confirmed a language with a WP.org language pack works.
* Confirmed a language relying on local translation (not translated enough for w.org pack) works.

PR #17, opened 2024-12-18
Github Action: Add phpcs checks

kraftbj:
Initial run at adding basic phpcs checks via GitHub Actions and Husky pre-commit hooks.

Testing: Create a PR into this branch with a PHP change. One change should be adding new code without error. One should be adding new code with an error. Another should be adding new code on a line with an existing error.

Test PR: https://github.com/kraftbj/secure-custom-fields/pull/1

### DIFF
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -21,11 +21,13 @@ jobs:
       run: composer install
         
     - name: Run PHPCS on changed files
+      env:
+        BASE_REF: ${{ github.base_ref }}
       run: |
-        CHANGED_FILES=$(git diff --name-only --diff-filter=d origin/${{ github.base_ref }} | grep ".php$" || true)
-        if [ ! -z "$CHANGED_FILES" ]; then
+        CHANGED_FILES=$(git diff --name-only --no-renames --diff-filter=d "origin/$BASE_REF" '*.php' || true)
+        if [[ -n "$CHANGED_FILES" ]]; then
           vendor/bin/phpcs-changed \
-            --git-base=origin/${{ github.base_ref }} \
+            --git-base="origin/$BASE_REF" \
             --phpcs-path=vendor/bin/phpcs \
             --standard=.phpcs.xml.dist \
             $CHANGED_FILES
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -1,0 +1,34 @@
+name: Code Standards
+on:
+  pull_request:
+
+jobs:
+  phpcs:
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0  # Required for phpcs-changed to compare with base branch
+        
+    - name: Setup PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: '8.1'
+        tools: composer:v2
+        
+    - name: Install Dependencies
+      run: composer install
+        
+    - name: Run PHPCS on changed files
+      run: |
+        CHANGED_FILES=$(git diff --name-only --diff-filter=d origin/${{ github.base_ref }} | grep ".php$" || true)
+        if [ ! -z "$CHANGED_FILES" ]; then
+          vendor/bin/phpcs-changed \
+            --git-base=origin/${{ github.base_ref }} \
+            --phpcs-path=vendor/bin/phpcs \
+            --standard=.phpcs.xml.dist \
+            $CHANGED_FILES
+        else
+          echo "No PHP files changed"
+        fi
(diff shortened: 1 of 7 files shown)


.github/workflows/phpcs.yml
```diff
@@ -0,0 +1,34 @@
...
+        CHANGED_FILES=$(git diff --name-only --diff-filter=d origin/${{ github.base_ref }} | grep ".php$" || true)
+        if [ ! -z "$CHANGED_FILES" ]; then
+          vendor/bin/phpcs-changed \
+            --git-base=origin/${{ github.base_ref }} \
+            --phpcs-path=vendor/bin/phpcs \
+            --standard=.phpcs.xml.dist \
+            $CHANGED_FILES
+        else
+          echo "No PHP files changed"
+        fi
```
anomiex: It's risky to substitute GitHub variables directly into shell scripts. Safer to use environment variables.

Also, you'll want to quote those environment veriables where they're used.

You may also want to pass `--no-renames` to `git diff`, otherwise it may sometimes get confused.

No need for a grep, just tell `git diff` to only do `.php` files directly.

Then, just a pet peeve, since it's bash we can use the built-in `[[ ]]` instead of `[ ]`.
```suggestion
      env:
        BASE_REF: ${{ github.base_ref }}
      run: |
        CHANGED_FILES=$(git diff --name-only --no-renames --diff-filter=d "origin/$BASE_REF" '*.php' || true)
        if [[ -n "$CHANGED_FILES" ]]; then
          vendor/bin/phpcs-changed \
            --git-base="origin/$BASE_REF" \
            --phpcs-path=vendor/bin/phpcs \
            --standard=.phpcs.xml.dist \
            $CHANGED_FILES
        else
          echo "No PHP files changed"
        fi
```

.husky/pre-commit-phpcbf.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/sh
+
+# Disable exit on error since we handle exit codes manually
+set +e
+
+# Get list of staged PHP files
+STAGED_PHP_FILES=$(git diff --cached --name-only --diff-filter=d | grep ".php$") || true
```
anomiex: If you're going to `set +e`, you can skip the `|| true` here.

Also you may want to sync this `git diff` with the earlier one.

.husky/pre-commit-phpcbf.sh
```diff
@@ -0,0 +1,53 @@
...
+
+echo "Running PHPCBF on staged files..."
+
+# First try to fix what we can with PHPCBF
+echo "$STAGED_PHP_FILES" | xargs vendor/bin/phpcbf
+
+echo 'does this run?'
+
+# Check the return code
+PHPCBF_STATUS=$?
```
anomiex, line 20: This is going to get the status of the `echo` on line 19, not the `xargs` on line 17. Move it up.

.husky/pre-commit-phpcbf.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# If files were fixed, add them back to staging
+if [ $PHPCBF_STATUS -ne 3 ]; then
+    echo "$STAGED_PHP_FILES" | xargs git add
+    
+    if [ $PHPCBF_STATUS -eq 1 ] || [ $PHPCBF_STATUS -eq 2 ]; then
+        echo "\nShowing remaining errors after auto-fixing:"
```
anomiex: The `\n` here probably won't be expanded.
```suggestion
        echo ""
        echo "Showing remaining errors after auto-fixing:"
```

.husky/pre-commit-phpcbf.sh
```diff
@@ -0,0 +1,53 @@
...
+        echo "$STAGED_PHP_FILES" | xargs vendor/bin/phpcs
+    fi
+fi
+
+# Now run phpcs-changed to check only modified lines
+echo "Checking modified lines with phpcs-changed..."
+STAGED_FILES_SPACE_SEPARATED=$(echo "$STAGED_PHP_FILES" | tr '\n' ' ')
```
anomiex: I don't think this is needed. The expansion below should be fine with newlines rather than spaces, even with a POSIX /bin/sh.

anomiex (commented):
Gave it a quick once-over.

.husky/pre-commit-phpcbf.sh
kraftbj: Fixed by removing the debug `echo` d1afd5f - thanks.

.husky/pre-commit-phpcbf.sh
kraftbj: 111a5b59897f64c21894db32791daf579a400cb4

.husky/pre-commit-phpcbf.sh
kraftbj: Seems to work without it. b2920566a68cd43832c5ee78b4335017eaeeb8cd

anomiex (approved):
Looks reasonable to me.

PR #28, opened 2025-02-05
Build Tools: Add NVM file and wp-env dev tools

cbravobernal:
## What

Adds a default `nvmrc`file, so developers can easily switch to a compatible version of Node to build the plugin.

It also adds a `wp-env` file, just in case developers can use that tool for development without any warning.


kraftbj (approved):
Looks good to me.

Issue #27, opened 2025-01-18
Old json in acf-json folder not showing up


haseebnawaz298:
in beta-4 they show and in beta-5 and latest they are not showing, if i sync then in beta-4 then update plugin they stay, i am unable to find the cause, why this is happening.

haseebnawaz298:
the sync also donot work in the latest version

kraftbj:
Are you still seeing this issue or have any errors in your site's error log? I'm not able to duplicate the issue so far.

haseebnawaz298:
https://www.awesomescreenshot.com/video/36487469?key=66dca7082c56bdec0e64cd34f5d16bb6

haseebnawaz298:
when can we get this change to be added to next version?

kraftbj:
I'll package up a new version on Tuesday in the U.S. morning. I was at and now traveling back from WC Asia; don't want to be unavailable on a plane if anyone finds anything unexpected once it has been released.

Thanks for your patience @haseebnawaz298!

PR #30, opened 2025-02-19
Fixes old json in acf-json folder not showing in Sync available

cyberwani:
https://github.com/WordPress/secure-custom-fields/issues/27


kraftbj (approved):
Thank you for the catch.

PR #44, opened 2025-02-28
Add nav menu field to scf

haseebnawaz298:
- this is a very common field that is used a lot, and it is not in scf.

### DIFF
--- a/docs/features/field/nav_menu/tutorial.md
+++ b/docs/features/field/nav_menu/tutorial.md
@@ -1,0 +1,18 @@
+# Using the Nav Menu Field
+
+## Basic Setup
+
+1. Create a new Field Group
+2. Add a Nav Menu field
+3. Configure options:
+   - Set Return Value
+   - Set Menu Container
+   - Set of Allow Null is (true/false)
+
+## Common Use Cases
+
+1. Use to output wordpress menu anywhere
+
+## Tips
+
+- to add more Menu Container use ``wp_nav_menu_container_allowed_tags`` hook
--- a/docs/features/field/nav_menu/tutorial.md
+++ b/docs/features/field/nav_menu/tutorial.md
@@ -6,7 +6,7 @@ To get started with the **Nav Menu Field** in SCF, follow these steps to configu
 
 ### **Step 1: Create a New Field Group**
 
-First, create a new field group in **Advanced Custom Fields (SCF)**. This field group will hold your custom fields, including the **Nav Menu** field.
+First, create a new field group in **Secure Custom Fields (SCF)**. This field group will hold your custom fields, including the **Nav Menu** field.
 
 1. Go to **Custom Fields > Add New** in your WordPress admin panel.
 2. Title your field group and choose the location rules for where this field group should be displayed (e.g., on a specific page or post type).
--- a/docs/features/field/nav_menu/tutorial.md
+++ b/docs/features/field/nav_menu/tutorial.md
@@ -11,7 +11,7 @@
 
 ## Common Use Cases
 
-1. Use to output wordpress menu anywhere
+1. Use to output WordPress navigation menu anywhere.
 
 ## Tips
 
--- a/docs/features/field/nav_menu/tutorial.md
+++ b/docs/features/field/nav_menu/tutorial.md
@@ -1,18 +1,71 @@
-# Using the Nav Menu Field
+# **Using the Nav Menu Field**
 
-## Basic Setup
+## **Basic Setup**
 
-1. Create a new Field Group
-2. Add a Nav Menu field
-3. Configure options:
-   - Set Return Value
-   - Set Menu Container
-   - Set of Allow Null is (true/false)
+To get started with the **Nav Menu Field** in SCF, follow these steps to configure and display a WordPress navigation menu:
 
-## Common Use Cases
+### **Step 1: Create a New Field Group**
 
-1. Use to output WordPress navigation menu anywhere.
+First, create a new field group in **Advanced Custom Fields (SCF)**. This field group will hold your custom fields, including the **Nav Menu** field.
 
-## Tips
+1. Go to **Custom Fields > Add New** in your WordPress admin panel.
+2. Title your field group and choose the location rules for where this field group should be displayed (e.g., on a specific page or post type).
 
-- to add more Menu Container use ``wp_nav_menu_container_allowed_tags`` hook
+### **Step 2: Add a Nav Menu Field**
+
+Once you've created a new field group, add a new field of type **Nav Menu**. This field will allow users to select a navigation menu from the available menus on your site.
+
+1. Click **Add Field** within the field group.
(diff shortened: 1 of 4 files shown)


includes/fields/class-acf-field-nav-menu.php
```diff
@@ -0,0 +1,207 @@
+<?php
+/**
+ * This is a PHP file containing the code for the acf_field_nav_menu class.
+ *
+ * @package wordpress/secure-custom-fields
+ * @since 6.4.1
```
kraftbj:
```suggestion
 * @since 6.5.0
```

includes/fields/class-acf-field-nav-menu.php
```diff
@@ -0,0 +1,207 @@
...
+	 * Acf Nav menu field class
+	 */
+	class Acf_Field_Nav_Menu extends acf_field {
+
+		/**
+		 * This function will setup the field type data
+		 *
+		 * @type    function
+		 * @date    5/03/2014
+		 * @since   ACF 6.4.1
```
kraftbj:
```suggestion
		 * @since  6.5.0
```

docs/features/field/nav_menu/tutorial.md
```diff
@@ -0,0 +1,18 @@
...
+1. Create a new Field Group
+2. Add a Nav Menu field
+3. Configure options:
+   - Set Return Value
+   - Set Menu Container
+   - Set of Allow Null is (true/false)
+
+## Common Use Cases
+
+1. Use to output wordpress menu anywhere
```
kraftbj:
```suggestion
1. Use to output WordPress navigation menu anywhere.
```

kraftbj (commented):
This works seems to work well. A couple of questions:

1. Themes without nav menus (e.g. Twenty Twenty Four) will output a notice on the post editor that the theme doesn't support it. I'm curious if there's a better way of handling it (notice on the field group editor? Disable the Nav Menu field when using a theme that doesn't support it? Hide the field on the post editor if the theme doesn't support it, etc.
2. Could you walk me through how you're using this? I'd just like to understand the user journey on it. Something like "this post is about Cats, so I want to display my Animals-specific navigation menu under the top nav" with more flexibility than, say, logic by category?

I mainly want to ensure I'm understanding the need properly.

There are a couple of version updates needed (this will go in 6.5) and tweaking the docs with a typo. [I'm happy to make those edits myself before merge, so no worries there].

With the tutorial, please feel free to be more verbose. The docs, overall, are really spartan but I hope to flesh them out so someone could follow the tutorial from start to finish to add a field, use a field via the post editor, and output the field.

haseebnawaz298:
1. we can go with showing a message in field group editor and hiding the field from post editor, then we donot need to show any notices.
2. I have worked on many sites where this was needed we did use a plugin for it, but that plugin is no longer maintained.

some of the common use case that i normally use is.

-  A second menu that can be set as a global (Theme Options) and also be edited in a page( Page options), if page option is set to be empty global one is fetched.
- it also gives you the ability to show different menus based on different conditions

haseebnawaz298:
I would also love to help with the docs, just need a direction on how deep we want the docs to be. eg: do we need to give code examples etc, or just basic instructions?

kraftbj:
> 1. we can go with showing a message in field group editor and hiding the field from post editor, then we donot need to show any notices.

Awesome; are you good with making those changes on this branch before merging? (note that I did commit the suggestions I made in the initial review so please pull before new commits)



> I would also love to help with the docs, just need a direction on how deep we want the docs to be. eg: do we need to give code examples etc, or just basic instructions?

I'd say basic code examples. I'm thinking of two types of users using the docs:
1. The experienced person looking for details. We don't need to do too much beyond good inline documentation of the args, describing the options, etc.
2. Someone new to WordPress/web development who need something more like training wheels.

For the latter person, if they said "I want to add a new field to my site" and stumbled upon SCF, will the tutorial provide them enough information to be able to add the field and output it? Until we have better integration with blocks, that's going to almost always be with small php code.

So, for the nav menu, I would say it could be something like "To output this field, add this code to the theme template, functions.php or elsewhere as needed for implementation `..... get_field(.....)......`"

Then with a small tutorial showing "As an example, using a child theme of the Twenty Twelve default theme, if your nav menu field is "special-menu", add [code] to the such and such template.php"

Does that make sense? Obviously few people would need the specific tutorial exactly as written, but enough where someone could get it work in that generic place and be enough to hopefully connect some dots for them.

Additionally, I am okay with merging this (with the notice update) without the docs being 100% and accepting a follow-up PR for it.

Thank you!

haseebnawaz298:
i will work on this hopefully today or tomorrow, will get back to you soon

haseebnawaz298:
This is how the notices show in the group field editor now,

field is hidden if there is no support in theme.

image
image

haseebnawaz298:
also updated the docs to be a bit better

includes/fields/class-acf-field-nav-menu.php
```diff
@@ -0,0 +1,237 @@
...
+		public function render_field_settings( $field ) {
+			$allow_null = $field['allow_null'];
+			$nav_menus  = wp_get_nav_menus( $allow_null );
+			if ( current_theme_supports( 'menus' ) ) {
+				if ( empty( $nav_menus ) ) {
+					?>
+					<div class="acf-field">
+						<div class="acf-notice">
+							<p>
+								<?php esc_html_e( 'Warning: You don\'t have any menus created please visit', 'secure-custom-fields' ); ?> <a href="<?php echo esc_url( admin_url( 'nav-menus.php' ) ); ?>"><?php esc_html_e( 'this', 'secure-custom-fields' ); ?></a> <?php esc_html_e( 'link to create menus.', 'secure-custom-fields' ); ?>
```
kraftbj: We'd probably want to incorporate the variable into the translation field: https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#variables

Additionally, can we have the actual link be more descriptive? Instead of "this", something like "This field requires an existing navigation menu, please ==create a menu first==." (`==` to show the linked text)

includes/fields/class-acf-field-nav-menu.php
```diff
@@ -0,0 +1,237 @@
...
+						</div>
+					</div>
+					<?php
+				}
+			} else {
+				?>
+				<div class="acf-field">
+					<div class="acf-notice">
+						<p>
+							<?php esc_html_e( 'Warning: The theme does not support navigation menus, the field will not show.', 'secure-custom-fields' ); ?>
```
kraftbj:
```suggestion
							<?php esc_html_e( 'Warning: The theme does not support navigation menus, the field will not display.', 'secure-custom-fields' ); ?>
```

kraftbj (changes requested):
I haven't actually tested it yet, but a couple of comments from reviewing the code.

kraftbj (approved):
Tests well.

haseebnawaz298:
nice

Issue #71, opened 2025-03-07
Code analysis: spelling errors


pkevan:
Many thousand mainly code comment level spellings exist within the repo, which is just untidy. Since the effort to change this is large, and it affects many files, unsure how to proceed aside from doing this piecemeal.

Is it worth the effort, or should it just be tackled adhoc?

pkevan:
further analysis reveals that a lot of these are through the translation files, so not quite as big an issue as initially thought.

kraftbj:
I think piecemeal is fine. I can be sure to merge those quickly so nothing grows stale.

priethor:
Could we consider this `Documentation` or `Code quality` rather than a bugfix? From experience in core and Gutenberg, anything that doesn't fix something _for the user or extender_ shouldn't be considered a bugfix.

kraftbj:
I'm good with that. I didn't see an option in GH for a custom issue types, but can go back to old fashioned labels.

pkevan:
Since the initial PRs have been merged, I think we're good to close this issue as resolved.

PR #77, opened 2025-03-10
Fix typos in misc files

pkevan:
See #71

### DIFF
--- a/docs/code-reference/META.md
+++ b/docs/code-reference/META.md
@@ -573,7 +573,7 @@ This file tracks code elements that need documentation.
 
 ## rest-api/class-acf-rest-api.php
 
-### Classs
+### Classes
 
 - `ACF_Rest_Api`
 
--- a/acf.php
+++ b/acf.php
@@ -13,6 +13,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 if ( is_admin() ) {
+	/** @phpstan-ignore-next-line */ // phpcs:ignore
 	include_once ABSPATH . 'wp-admin/includes/plugin.php';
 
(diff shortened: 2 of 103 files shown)


kraftbj:
The `code-reference` section is auto-generated from the source code, so most of these would be auto-corrected once the source docblocks are updated. I think the META.md is the one that will need to be fixed in the generator itself.

docs/code-reference/META.md
```diff
@@ -573,7 +573,7 @@ This file tracks code elements that need documentation.
 
 ## rest-api/class-acf-rest-api.php
 
-### Classs
+### Classes
```
kraftbj, line 576: We to update this in the `docs/bin/generate-parsed-md.php` file, line ~544 to:
```
				if ( 'class' === $type ) {
					$type = 'classes';
				}
```

kraftbj (changes requested):

docs/code-reference/META.md
pkevan: 👍

pkevan: ah! i understand now, it's just pluralising the names

pkevan: See: https://github.com/WordPress/secure-custom-fields/pull/77/commits/de86a3f0f111228ea2d578276d6f5ab95d03aa50

pkevan:
> The `code-reference` section is auto-generated from the source code, so most of these would be auto-corrected once the source docblocks are updated. I think the META.md is the one that will need to be fixed in the generator itself.

I'll regenerate these, and see what changes get resolved.

pkevan:
OK - think that should be it. Thanks for the pointers @kraftbj 😄

kraftbj (approved):

Issue #67, opened 2025-03-04
PHPStan: correct includes/forms/**


kraftbj:
```
			message: '#^Access to an undefined property acf_form_customizer\:\:\$preview_errors\.$#'
			identifier: property.notFound
			count: 1
			path: ../includes/forms/form-customizer.php

		-
			message: '#^Access to an undefined property acf_form_customizer\:\:\$preview_fields\.$#'
			identifier: property.notFound
			count: 2
			path: ../includes/forms/form-customizer.php

		-
			message: '#^Access to an undefined property acf_form_customizer\:\:\$preview_values\.$#'
			identifier: property.notFound
			count: 2
			path: ../includes/forms/form-customizer.php

		-
			message: '#^Action callback returns type but should not return anything\.$#'
			identifier: return.void
			count: 1
			path: ../includes/forms/form-nav-menu.php

		-
			message: '#^Method acf_form_nav_menu\:\:wp_nav_menu_item_custom_fields\(\) should return type but return statement is missing\.$#'
			identifier: return.missing
			count: 2
			path: ../includes/forms/form-nav-menu.php

		-
			message: '#^Action callback returns int but should not return anything\.$#'
			identifier: return.void
			count: 1
			path: ../includes/forms/form-post.php

		-
			message: '#^Access to an undefined property acf_form_widget\:\:\$preview_errors\.$#'
			identifier: property.notFound
			count: 1
			path: ../includes/forms/form-widget.php

		-
			message: '#^Access to an undefined property acf_form_widget\:\:\$preview_reference\.$#'
			identifier: property.notFound
			count: 1
			path: ../includes/forms/form-widget.php

		-
			message: '#^Access to an undefined property acf_form_widget\:\:\$preview_values\.$#'
			identifier: property.notFound
			count: 1
			path: ../includes/forms/form-widget.php
```

pkevan:
Closed via: https://github.com/WordPress/secure-custom-fields/pull/122

PR #81, opened 2025-03-13
Form Post: Fix PHPStan and PHPCS issues.

cbravobernal:
## What

Part of #67

Refactor `form-post.php` file to follow PHPCS standards and PHPStan issues. I did manual testing and everything seems to be ok, but I would feel more comfortable if we plan some kind of unit or e2e testing.

The riskiest change is removing `$postarr` from `wp_insert_post_empty_content`. However the variable is not being used inside the function.

### DIFF
--- a/includes/forms/form-post.php
+++ b/includes/forms/form-post.php
@@ -187,7 +187,7 @@ if ( ! class_exists( 'ACF_Form_Post' ) ) :
 		public function edit_form_after_title() {
 
 			// globals
-			global $post, $wp_meta_boxes;
+			global $post;
 
 			// render post data
 			acf_form_data(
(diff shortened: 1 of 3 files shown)


includes/forms/form-post.php
```diff
@@ -6,10 +6,22 @@
 
 if ( ! class_exists( 'ACF_Form_Post' ) ) :
 
+	/**
+	 * Class ACF_Form_Post
+	 *
+	 * Handles the functionality for adding custom fields to post edit screens.
+	 *
+	 * @package ACF
+	 * @since 5.0.0
```
kraftbj:
```suggestion
	 * @since ACF 5.0.0
```

tests/php/bootstrap.php
```diff
@@ -13,3 +13,6 @@
 
 // Load our plugin.
 require dirname( dirname( __DIR__ ) ) . '/secure-custom-fields.php';
+
+// Load the ACF_Form_Post class.
+require dirname( dirname( __DIR__ ) ) . '/includes/forms/form-post.php';
```
kraftbj: Can you share your thought on adding it to the main bootstrap? My first thought without digging into anything would be to have this require within the test-form-post.php file.

includes/forms/form-post.php
cbravobernal: Done!

tests/php/bootstrap.php
cbravobernal: I required the file on the test one, as you mentioned. Honestly, I'm not 100% sure of the best approach for these unit tests. Let's see in the future for functions that require several files.

kraftbj (approved):

PR #87, opened 2025-03-25
Playground: Add blueprint json asset file

cbravobernal:
## What

Adds `blueprint.json` file to enable live preview in plugins.

Followed this doc: https://developer.wordpress.org/plugins/wordpress-org/previews-and-blueprints/#enabling-plugin-previews

Co-authored with @santosguillamot.

PD: I may need extra permissions to test this, a plugin committer needs to enable live previews 😅

### DIFF
--- a/assets/blueprints/blueprint.json
+++ b/assets/blueprints/blueprint.json
@@ -1,0 +1,23 @@
+{
+  "landingPage": "/wp-admin/edit.php?post_type=acf-field-group",
+  "features": {
+    "networking": true
+  },
+  "steps": [
+    {
+        "step": "login",
+        "username": "admin",
+        "password": "password"
+      },
+    {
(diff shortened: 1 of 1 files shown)


kraftbj:
Let's give it a go. The blueprint file needs to be present in order to activate live preview. I'll update trunk svn to see if that's enough to allow for live preview.

kraftbj:
It's alive! I was too eager in approving—the file is in the wrong location. Resolving it in #90.

cbravobernal:
> It's alive! I was too eager in approving—the file is in the wrong location. Resolving it in #90.

First time doing this 😅

kraftbj:
@cbravobernal Same here! It's a great addition to the Plugin Directory but most of the plugins I touch day-to-day aren't really suitable to use it!

PR #92, opened 2025-03-26
Legacy: begin a legacy mode with swapping of hooks (DNM)

kraftbj:
As time evolves, SCF is likely to diverge from the ACF parent. Similar to the great renaming in WordPress, SCF should start to move things to a new naming scheme. For backwards compatiblity, I'm thinking of having a "legacy mode". Open to thoughts on the idea in principle and in implementation!

Additionally, need to confirm `scf_` is the best prefix moving forward.


aminul-xs:
When will these pull requests be merged?

kraftbj:
Closing this until any roadmap is determine off of #150

Issue #85, opened 2025-03-20
Shortcode Bug


YanMetelitsa:
Greetings!

I noticed, that on some sites the `[acf]` shortcode does not work, while on some sites it works (with the same versions of WordPress and SCF).

When trying to output shortcode `[acf field=”..."]`, an empty string is displayed. Everywhere – in page/post content, plugins etc..

I checked if the shortcode is registered using $GLOBALS[ 'shortcode_tags' ] – shortcode `[acf]` is registered.

At the same time, if you recreate the shortcode, it starts working correctly:

```
add_shortcode( 'acf', function ( $atts ) {
	$value = apply_filters(
		'acf/format_value', 
		get_field( $atts[ 'field' ], $atts[ 'post_id' ] )
	);

	return $value;
});
```

pkevan:
Hi @YanMetelitsa, thanks for the report - is there any further details you can share with the sites you were testing this on? and any differences between the ones which work and don't i.e. themes/plugins?

YanMetelitsa:
Hi, @pkevan! The sites are practically the same. I dug a bit deeper into the issue and found out:

- Starting from ACF 6.3, shortcode is disabled by default in new installations for security reasons.
- On the site where the shortcode works, ACF was installed first, then SCF was added (system status shows it as enabled).
- On the site where the shortcode doesn't work, SCF was installed right away (system status shows it as disabled).
- I added the shortcode activation hook to the second site:
```
add_action( 'acf/init', function () {
    acf_update_setting( 'enable_shortcode', true );
});
```
- Now the system status shows the shortcode as enabled.
- The shortcode still doesn't work.

That's all the information I've managed to find out so far.

pkevan:
Thanks - I can see it here in the code: https://github.com/WordPress/secure-custom-fields/blob/trunk/secure-custom-fields.php#L172-L175 but don't fully understand if or why or when it should be possible to do this - i'll investigate further.

YanMetelitsa:
Okay, I'll also try to find time to look for the cause of this error in the code. Especially since I've already been contacted several times with this problem, that is, it's quite widespread and, in theory, occurs on any new SCF installation.

pkevan:
[...]

PR #95, opened 2025-03-28
Fix empty shortcode output

YanMetelitsa:
Fixies Issue #85 (Shortcode Bug)

Perhaps the text of the message needs to be changed as I am not a native speaker.

### DIFF
--- a/includes/api/api-template.php
+++ b/includes/api/api-template.php
@@ -1084,7 +1084,8 @@ function acf_shortcode( $atts ) {
 		if ( is_preview() ) {
 			return apply_filters( 'acf/shortcode/field_not_allowed_message', '[' . esc_html__( 'The requested ACF field is not allowed to be output in bindings or the ACF Shortcode.', 'secure-custom-fields' ) . ']' );
 		} else {
-			return apply_filters( 'acf/shortcode/field_not_allowed_message', '[' . sprintf( esc_html__( 'You need to enable "Allow Access to Value in Editor UI" in the "%s" field parameters to display it in the ACF Shortcode.', 'secure-custom-fields' ), $field[ 'label' ] ) . ']' );
+			/* translators: %s Field name */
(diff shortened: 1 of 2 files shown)


kraftbj:
@pkevan Since you dug into the original report a bit, would you be able to review this one?

pkevan (commented):
I think we need to be careful with outputting messages to end users here (maybe an error warning which doesn't show in the front-end would be better).

YanMetelitsa:
@pkevan, in any case, in my opinion, it is necessary to make some kind of display of an explanation of why the shortcode "does not work". In my experience, this raises questions and bewilderment among some users.

kraftbj:
What about having the message only displayed for logged-in admins?

YanMetelitsa:
> What about having the message only displayed for logged-in admins?

After my last post I thought the same thing - display for admin, something like that:

```
...
else {
	if ( current_user_can( 'manage_options' ) ) {
		/* translators: %s Field name */
		return apply_filters( 'acf/shortcode/field_not_allowed_message', '[' . sprintf( esc_html__( 'You need to enable "Allow Access to Value in Editor UI" in the "%s" field parameters to display it in the ACF Shortcode.', 'secure-custom-fields' ), $field['name'] ) . ']' );
	}
	
	return;
}
...
```

YanMetelitsa:
Also, my bad, I think it would be safer to organize the output as follows: 

`apply_filter( 'filter', esc_html( '[' . sprintf( __( '%s', 'secure-custom-fields' ), $field ) . ']' ) );`

pkevan:
> What about having the message only displayed for logged-in admins?

That could work - or even have that cap check filtered? As it might be better to have editors/admins included by default (on the basis on advanced privileges) 

Given there are multiple instances of outputting based on which settings are set outside of the shortcode, we should probably cover all of those which return an empty response.

priethor:
Hi @pkevan @YanMetelitsa, is anybody actively working on this?

pkevan:
Not at present for me - i'll unassign myself.

YanMetelitsa:
> Hi @pkevan @YanMetelitsa, is anybody actively working on this?

Hi!

I thought this problem was solved already, but I can't say for sure, it needs to be checked.

pkevan:
Yes, was resolved in #114.

PR #96, opened 2025-03-31
E2E: Create a field group with a text input.

cbravobernal:
It's still a draft. I need to cleanup after the tests and see if we want a constants file for e2e testing.


cbravobernal:
Closing in favor of #97

PR #97, opened 2025-03-31
E2E: Refactor to use wp-scripts library.

cbravobernal:
## What

- Use wp-scripts default e2e testing suite to be closer to Core development procedures.


cbravobernal:
Closed in favor of #98

PR #99, opened 2025-04-01
Add composer installation to docs/welcome

racmanuel:
Hello, I have added how to install Secure Custom Fields in a plugin or theme using composer.


docs/welcome/installation.md
```diff
@@ -27,6 +27,76 @@ Before installing, ensure your site meets these requirements:
...
+⚠️ **Note:** Replace MY_SCF_PATH and MY_SCF_URL with constants that match your plugin/theme structure if necessary.
+
+### Done!
+You have successfully installed and integrated Secure Custom Fields via Composer. You can now use it as you would with a normal installation, but with all the benefits of Composer-based dependency management.
+
+### Hide SCF Admin Menu and Updates
+
+```php
+// Hide the SCF admin menu item.
+add_filter( 'acf/settings/show_admin', '__return_false' );
```
priethor, line 130: We should be careful not to suggest always disabling the admin menu: what if the site using the plugin/theme uses SCF anyway?

docs/welcome/installation.md
```diff
@@ -27,6 +27,76 @@ Before installing, ensure your site meets these requirements:
 3. Upload the plugin folder to `/wp-content/plugins/`
 4. Activate through the WordPress admin interface
 
+### Composer Installation
+
+This guide explains how to install and integrate the **Secure Custom Fields** plugin in your WordPress theme or plugin using Composer.
```
priethor: Could you add a bit more about the `why`? A one liner with an example should be enough.

priethor (changes requested):
Thanks for submitting this! I've left a couple of notes.

docs/welcome/installation.md
racmanuel: @priethor You're right, I think this should be an optional step, I'll modify it in a new commit.

docs/welcome/installation.md
racmanuel: @priethor I made in the new commit and pr para añadir una section about : Why integrate Secure Custom Fields (SCF) with Composer.

docs/welcome/installation.md
priethor: The same goes for plugin updates. I think it should be checked whether SCF is already installed with `get_plugins()`  before trying to hide either.

docs/welcome/installation.md
```diff
@@ -26,11 +26,132 @@ Before installing, ensure your site meets these requirements:
...
+      "url": "https://wpackagist.org"
+    }
+  ],
+  "extra": {
+    "installer-paths": {
+      "vendor/{$name}/": ["type:wordpress-plugin"]
+    }
+  },
+  "require": {
+    "wpackagist-plugin/secure-custom-fields": "6.4.1"
```
priethor:
```suggestion
    "wpackagist-plugin/secure-custom-fields": "^6.4"
```

Otherwise, it will always install that specific version; let's not take for granted that readers are familiar with Composer 🙂

docs/welcome/installation.md
```diff
@@ -26,11 +26,132 @@ Before installing, ensure your site meets these requirements:
...
+
+- **Hide Update Notifications:**  
+  The second filter disables the SCF update notices, so users won't see update prompts for the plugin inside the admin dashboard.
+
+#### When should you use it?
+
+- If you are bundling SCF inside your plugin or theme and want to **control all the custom fields yourself** without allowing clients or users to modify them.
+- If you want to **maintain full control** over SCF versions and updates to avoid compatibility issues caused by manual updates.
+
+> **Note:** Hiding updates means you are responsible for manually updating SCF when necessary to keep your project secure and compatible.
```
priethor:
```suggestion
> **Note:** Hiding updates means you are responsible for manually updating SCF when necessary to keep your project secure and compatible, but it also helps avoid potential conflicts between Composer and the built-in updater.
```

priethor (approved):

priethor (approved):

PR #101, opened 2025-04-02
Revert text domain change

pkevan:
In order for the filter to be named correctly (https://developer.wordpress.org/reference/functions/shortcode_atts/), the 3rd parameter should be the shortcode.


cbravobernal (approved):
LGTM

PR #102, opened 2025-04-02
E2E: Add a trash all post types step.

cbravobernal:
## What

Adds a `trashAllPostTypes` helper to clean the testing environment when you run it locally. On GitHub Actions, the env is destroyed after running it, but still cleaning is a good approach to keep consistency with future tests.


pkevan (approved):
👍

PR #103, opened 2025-04-04
E2E: Add field text testing.

cbravobernal:
## What

Includes a test and the boilerplate for a text field creation, cleaning, and frontend rendering check.

### DIFF
--- a/tests/e2e/plugins/scf-test-get-field.php
+++ b/tests/e2e/plugins/scf-test-get-field.php
@@ -7,8 +7,6 @@
  * @package scf-test-plugins
  */
 
-add_filter( 'the_content', 'scf_add_get_field_at_the_end' );
-
 /**
  * Add post-formats support to pages
  */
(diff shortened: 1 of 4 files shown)


tests/e2e/plugins/scf-test-get-field.php
```diff
@@ -0,0 +1,39 @@
...
+		$field = is_string( $field ) ? $field : '';
+
+		// Sanitize the field value using WordPress sanitization functions.
+		$field = sanitize_text_field( $field );
+
+		// Escape the output for HTML context.
+		$escaped_field = esc_html( $field );
+
+		// Use wp_kses_post to allow safe HTML if needed, but escape by default.
+		$output = wp_kses_post( '<br><p id="scf-test-movie-title">Movie title: ' . $escaped_field . '</p>' );
```
cbravobernal: @pkevan Should SCF escape automatically within the get_field function?

Docs says that you need to do this `$escaped_wysiwyg = get_field('wysiwyg', false, true, true);`, but also recommends using `wp_kses_post`.

Am I being redundant here? I have that feeling.

cbravobernal:
Let's add a cleaning step before every test to avoid the errors we have right now.

tests/e2e/plugins/scf-test-get-field.php
gziolo: `wp_kses_post` filters out unallowed tags and attributes. In effect, if you would not escape the field and the HTML tags are allowed they would get printed inside the paragraph. So it sounds like a decision of whether you allow any HTML for the field here.

pkevan: > @pkevan Should SCF escape automatically within the get_field function?

Generally this isn't expected, and would more than likely get flagged when using phpcs anyway. 

Escaping should be in the realm of whatever is outputting the data, so in this case it's not needed, but generally any outputting functions within the plugin which requires no user input should be escaping it after using `get_field` i.e. the shortcodes.

cbravobernal: This one does not require user input cause the input will be only filled in an automated test. So I guess we are fine with removing it then.

PR #104, opened 2025-04-07
Refactor: Prevent some `undefined` array key warnings.

cbravobernal:
## What

While working on SCF. I have been receiving several warnings that made field creation difficult. Something like this:

Screenshot 2025-04-07 at 18 10 13
Screenshot 2025-04-07 at 18 10 19

## Test

- Check that you have similar warnings than the ones shown in the screenshot when you try to create custom fields.
- Check that with the PR they are gone.

### DIFF
--- a/includes/fields/class-acf-field-checkbox.php
+++ b/includes/fields/class-acf-field-checkbox.php
@@ -70,22 +70,23 @@ if ( ! class_exists( 'acf_field_checkbox' ) ) :
 			);
 
 			// append to class
-			$ul['class'] .= ' ' . ( isset( $field['layout'] ) && 'horizontal' === $field['layout'] ? 'acf-hl' : 'acf-bl' );
-			$ul['class'] .= ' ' . ( isset( $field['class'] ) ? $field['class'] : '' );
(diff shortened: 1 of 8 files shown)


includes/fields/class-acf-field-checkbox.php
```diff
@@ -189,31 +191,34 @@ function render_field_custom( $field ) {
...
+						'name'  => isset( $field['name'] ) ? $field['name'] : '',
+						'value' => $value,
+					);
 
-				// bail early if choice already exists
-				if ( in_array( $esc_value, $this->_values ) ) {
-					continue;
-				}
+					// bail early if choice already exists
+					if ( in_array( $esc_value, $this->_values, true ) ) {
```
pkevan, line 214: was this intentionally changed to a strict comparison?

pkevan:
I wonder if it might be best to have a helper function for all these `isset` checks, as it makes the code harder to read as a result?

includes/fields/class-acf-field-checkbox.php
cbravobernal: Yep, otherwise phpcs was complaining about it. It should not be risky. As both values come from `esc_attr`functions, which should return a string.

cbravobernal:
> I wonder if it might be best to have a helper function for all these `isset` checks, as it makes the code harder to read as a result?

I found indeed an `acf_maybe_get` function.

PR #109, opened 2025-04-14
Languages: es_EC is now served via w.org packs

kraftbj:
Per https://translate.wordpress.org/projects/wp-plugins/secure-custom-fields/language-packs/ , es_EC is now above threshold and is served via WordPress.org language packs.


kraftbj:
#110 would automate this process.

PR #110, opened 2025-04-14
Translations: add script to remove 90% translations

### LINKED ISSUE #105: Language pack removal
In https://github.com/WordPress/secure-custom-fields/pull/23 there were two language packs removed from `/lang` due to being a high enough coverage on https://translate.wordpress.org/projects/wp-plugins/secure-custom-fields/, but it's not clear what that percentage is. We should document what level we feel is a suitable level, and remove as necessary.

kraftbj:
Fixes #105

When preparing for a new release via `composer prepare-release` or a new `composer update-translations`, check for language packs on WordPress.org and remove any that are at 90%.

Future work would be to see if we can programmatically export the current translate.wordpress.org files for languages so any improvements made by volunteers make it to site owners.

### DIFF
--- a/bin/prepare-release.php
+++ b/bin/prepare-release.php
@@ -49,6 +49,7 @@ class Release_Preparation {
 		$this->build_assets();
 		$this->run_tests();
 		$this->generate_docs();
+		$this->update_translations();
 		$this->commit_changes();
 
 		$current_version = $this->get_current_version();
@@ -186,7 +187,7 @@ class Release_Preparation {
 			if ( 0 !== $return ) {
 				exit( $return );
 			}
-			passthru( 'git commit -m "Build assets and documentation"', $return );
+			passthru( 'git commit -m "Build assets, documentation, and translations"', $return );
 			if ( 0 !== $return ) {
 				exit( $return );
 			}
@@ -345,6 +346,15 @@ class Release_Preparation {
 		$line   = strtolower( trim( fgets( $handle ) ) );
 		fclose( $handle );
 		return 'y' === $line;
+	}
+
+	/**
+	 * Update plugin translations and cleanup
+	 */
+	private function update_translations() {
+		require_once __DIR__ . '/update-translations.php';
+		$updater = new Translation_Updater();
+		$updater->run();
 	}
 }
 
--- a/bin/prepare-release.php
+++ b/bin/prepare-release.php
@@ -50,6 +50,8 @@ class Release_Preparation {
 		$this->run_tests();
 		$this->generate_docs();
 		$this->update_translations();
+		// need to return to correct directory after running `update_translations`.
+		exec( chdir( dirname( __DIR__ ) ) );
 		$this->commit_changes();
 
 		$current_version = $this->get_current_version();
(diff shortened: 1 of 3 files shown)


kraftbj:
This can be tested by running the command and seeing `es_EC` removed until #109 is merged.

pkevan:
I wonder if we want to have a set baseline of translated languages? i.e. what happens when a language doesn't hit the 90% threshold after a major update and there is no longer a pack?

kraftbj:
Once there is a language pack, it'll continue to get generated, even if the translation percentage drops below the threshold ( https://github.com/WordPress/wordpress.org/blob/7bcd368b66c9e098bce1314a9f4af61d7ac067a1/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-language-pack.php#L649 ).

Once a language meets the threshold, we don't need to worry about it from the code/repo POV. On the polyglot side, we can aim to have PTE (project translation editors), which are folks who can approve a plugin's strings, so we can better ensure new and updated strings are translated.

There were some conversations/proposals about not serving a pack if the percentage dropped below a number, but those weren't accepted. (e.g. https://make.wordpress.org/polyglots/2020/03/10/delete-projects-language-pack-when-below-specific-percentage/ )

It would make sense to me to try to target the languages we can with more than 1% of WordPress installs (looking at https://wordpress.org/stats/ but open to other ways to determine higher-impact languages) which current lists:

Those we lack packs for:
Japanese | 5.83%
German | 5.49%
French (France) | 4.63%
Portuguese (Brazil) | 3.87%
English (UK) | 3.31% -- arguably less critical
Italian | 3.18%
Polish | 2.06% 
Turkish | 1.08%
Persian | 1.07%
Others | 3.25%

Those we do:
Spanish (Spain) | 5.81% -- plus a couple of variants
Dutch | 2.37% -- both Dutch and Dutch Belgium
Russian | 2.32%
Vietnamese | 1.07%

Something I'd like to do in a follow-up is see if we can programmatically pull down the translations from translate.wordpress.org for those not at the threshold yet so those that we do have in the repo are up to date. Right now, the files we serve have already been uploaded to translate.wordpress.org and, in theory, added/improved, but users aren't seeing those improved translations until we export them back into the repo (which I haven't done thus far)

pkevan (approved):
Tested with `composer update-translations` which works as expected.

Hitting some issues with `composer prepare-release` - will followup separately.

pkevan:
> Hitting some issues with `composer prepare-release` - will followup separately.

Added a [commit](https://github.com/WordPress/secure-custom-fields/pull/110/commits/8bb09f78b0f7fc13f7f967127cf89657a05c6600) to resolve one of the issues, needs testing.

kraftbj:
@pkevan Thanks for catching that. In my iterations on the script, I ended up not needing the `chdir` that was throwing things off. Can you give this a gander to confirm it works without needing the special chdir within the chain of commands?

pkevan:
> @pkevan Thanks for catching that. In my iterations on the script, I ended up not needing the `chdir` that was throwing things off. Can you give this a gander to confirm it works without needing the special chdir within the chain of commands?

Yes, will do and report back.

pkevan:
Looks good now @kraftbj 👍

PR #113, opened 2025-04-22
Framework and admin screen to add new features as opt-in

cbravobernal:
This PRs sets a standard for creating experiments and checking if they are enabled in JavaScript.

As there are no experiments ready yet, I commented all the code that initializes them with a `// Temporarily disabled` comment.

The PR was getting big, so it is better to land this first and work later on next experiments.

If you want to test this out, just uncomment the lines // Temporarily disabled in `admin-experiments.php` file and you will have something like this:

Screenshot 2025-04-25 at 21 44 31

Screenshot 2025-04-25 at 21 44 47


includes/admin/admin-experiments.php
```diff
@@ -0,0 +1,342 @@
...
+		/**
+		 * This function will setup the class functionality
+		 *
+		 * @since   SCF 6.4.2
+		 *
+		 * @return  void
+		 */
+		public function __construct() {
+			// Temporarily disabled - will be enabled when experiments feature is ready
+			// add_action( 'admin_menu', array( $this, 'admin_menu' ), 20 );
```
priethor: We should check whether $this->experiments is empty to avoid adding the menu if there are no experiments, except for testing purposes.

assets/src/js/README.md
```diff
@@ -0,0 +1,56 @@
...
+    // Experiment is disabled, use default behavior
+  }
+}
+```
+
+### Adding New Experiments
+
+To add a new experiment:
+
+1. Create a new experiment class in `includes/admin/experiments/`
```
priethor: Is this a "good practices" recommendation or a requirement? What's the reason to create the experiments in this folder instead of creating them in the place they belong, and just registering?

assets/src/js/_acf-experiments.ts
```diff
@@ -0,0 +1,63 @@
+/**
+ * Experiments TypeScript
+ *
+ * @package    Secure Custom Fields
+ * @subpackage Admin
+ * @since      6.4.2
```
priethor: 6.4.2 is out, shouldn't it be 6.5?

assets/src/js/_acf-experiments.ts
```diff
@@ -0,0 +1,63 @@
...
+      };
+    };
+  }
+}
+
+// Create a module to avoid global scope augmentation issues
+export {};
+
+// Initialize the SCF namespace if it doesn't exist
+if (typeof window.SCF === 'undefined') {
```
priethor: Shouldn't we reuse the existing `acf` namespace until we migrate it?

priethor (commented):
Thanks, Carlos, this is looking good! Since this is still a draft I didn't go too deep, but still left some early feedback ☺️

assets/src/js/README.md
cbravobernal: Ideally, this way it should be easier to find all experiments registered instead of having to dig into the entire codebase. Also, I'm trying to follow the PHP class based plugin structure, which may differ from how we did things in Gutenberg.

Is more a recommendation than a requirement, but this way all the admin interface will be automatically generated on PHP.

assets/src/js/_acf-experiments.ts
cbravobernal: I did this before the update 😅

assets/src/js/_acf-experiments.ts
cbravobernal: Yep. You are right. Otherwise it will be difficult to differentiate which is inside of each object.

includes/admin/admin-experiments.php
```diff
@@ -0,0 +1,342 @@
+<?php // phpcs:disable Universal.Files.SeparateFunctionsFromOO.Mixed
```
cbravobernal, line 1: Remove this.

gziolo:
It would be helpful to extract all changes related to code formatting to another PR and land separately, as it makes the review process more complex. What is the reason GitHub reports so massive changes in the lock file?

cbravobernal:
> It would be helpful to extract all changes related to code formatting to another PR and land separately, as it makes the review process more complex.

Sure, done in https://github.com/WordPress/secure-custom-fields/pull/115

> What is the reason GitHub reports so massive changes in the lock file?

It seems it was due to the update of wp-scripts and all its dependencies. I will take a look at it.

assets/src/js/README.md
cbravobernal: Let's remove the readme to prevent external devs to create more experiments.

includes/admin/admin-experiments.php
cbravobernal: Fetching the experiments on admin_menu action seems to be tricky, as the class handling all the stuff is being processed later. I'll added a much simpler approach with a message of "no experiments" in the admin view.

If we see there will be a release with no experiments, then we can comment just the first experiments call and will save some processing.

Screenshot 2025-04-25 at 10 27 54

gziolo: What's the target audience for these experiments? If it's only for development purposes then exposing the menu item behind `SCRIPT_DEBUG` flag should do the trick.

includes/admin/views/global/navigation.php
```diff
@@ -173,8 +173,10 @@ function acf_print_menu_section( $menu_items, $section = '' ) {
 $screen = get_current_screen();
 
 if ( ! in_array( $screen->id, acf_get_internal_post_types(), true ) ) {
-	if ( $plugin_page == 'acf-tools' ) {
+	if ( 'acf-tools' === $plugin_page ) {
 		$acf_page_title = __( 'Tools', 'secure-custom-fields' );
+	} elseif ( 'scf-experiments' === $plugin_page ) {
+		$acf_page_title = __( 'Beta Features', 'secure-custom-fields' );
```
gziolo: There is some inconsistency in the naming here. What's the difference between beta and experimental features?

includes/admin/admin-experiments.php
```diff
@@ -0,0 +1,328 @@
...
+
+			$this->check_submit();
+
+			acf_enqueue_scripts();
+
+			// Temporarily disabled - Uncomment to enable in production when ready.
+			// acf_enqueue_script( 'acf-experiments' );
+
+			// Temporarily disabled - Localize experiments data. Uncomment to enable in production when ready.
+			// $this->enqueue_experiments_script();
```
gziolo: What's the relationship between the scripts? Why not to put all of them inside `enqueue_experiments_script` method?

includes/admin/views/experiments/experiments.php
```diff
@@ -0,0 +1,111 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+// Get current screen and experiments.
+$screen      = get_current_screen();
+$experiments = acf()->admin_experiments->get_experiments();
+?>
+<div class="wrap" id="scf-admin-experiments">
+	<h1 class="wp-heading-inline"><?php esc_html_e( 'Experiments', 'secure-custom-fields' ); ?></h1>
```
gziolo: I really think we should land on one naming schema and call it `Experiments` or `Beta features` consistently. This is what I see at the moment:

Screenshot 2025-04-25 at 11 38 50

includes/admin/views/experiments/experiments.php
```diff
@@ -0,0 +1,111 @@
...
+		<div class="scf-experiments-header">
+			<p><?php esc_html_e( 'Enable or disable beta features. These features are in development and may change in future releases.', 'secure-custom-fields' ); ?></p>
+		</div>
+		
+		<?php if ( empty( $experiments ) ) : ?>
+			<div class="scf-no-experiments">
+				<p><?php esc_html_e( 'No beta features are currently available.', 'secure-custom-fields' ); ?></p>
+			</div>
+		<?php else : ?>
+		<form method="post" action="">
```
gziolo: Action is missing here.

assets/src/js/experiments/index.ts
```diff
@@ -0,0 +1,54 @@
...
+	interface Window {
+		acf: ACF;
+		acfExperiments?: ExperimentsData; // Variable created by wp_localize_script in admin-experiments.php L150.
+	}
+	var acf: ACF;
+	var acfExperiments: ExperimentsData | undefined;
+}
+
+// Create a module to avoid global scope augmentation issues.
+export {};
```
gziolo: Can you elaborate on why export is necessary here?

includes/admin/experiments/class-scf-admin-experiment-editor-sidebar.php
```diff
@@ -0,0 +1,52 @@
...
+
+		/**
+		 * Initialize the experiment.
+		 *
+		 * @return void
+		 */
+		public function initialize() {
+			$this->name        = 'editor-sidebar';
+			$this->title       = __( 'Move Elements to Editor Sidebar', 'secure-custom-fields' );
+			$this->description = __( 'Moves field group elements to the editor sidebar for a cleaner interface.', 'secure-custom-fields' );
```
gziolo: This will need some better phrasing, as it's unclear what elements it refers to and in what context. It also should be less about cleaner interface but more about the fact that custom fields can be managed using the same interface as any possible plugin integrated with the editor.

gziolo:
I can confirm that experiments get enabled or disabled with the checkbox after I uncommented the temporiraly commented out code. Unfortunately, I couldn't test more as the feature included doesn't have any logic yet.

I left some feedback, but it mostly refers to user facing conisderations and the way code is structured. Nothing really that would be concerning. Everything is on track to landing this PR.

cbravobernal:
> I can confirm that experiments get enabled or disabled with the checkbox after I uncommented the temporiraly commented out code. Unfortunately, I couldn't test more as the feature included doesn't have any logic yet.
> 
> I left some feedback, but it mostly refers to user facing considerations and the way code is structured. Nothing really that would be concerning. Everything is on track to landing this PR.

I will address all the concerns. Many of them has been due to the decision of renaming from experiments to beta features, and is much better to start with a good base rather than having to refactor in a future.

includes/admin/views/experiments/experiments.php
priethor: Agreed, let's go with Beta Features consistently.

includes/admin/admin-experiments.php
priethor: End users should have access to these experiments. It's more about opting in to new and usable features (e.g., new DataViews screens) than exposing half-baked features.

assets/src/js/experiments/index.ts
cbravobernal: If a file does not have any import or export, it is considered a script and its declarations (like variables, types, interfaces) are added to the global scope.

It is true that we are working with the global acf file. But don't think we want the interfaces to be global yet.

includes/admin/admin-experiments.php
cbravobernal: That's it, I exposed this one so more people can use the "experiments" framework to add their own. Ideally, we should not merge any feature that is not stable as a minimum.

includes/admin/admin-experiments.php
cbravobernal: `enqueue_experiments_script` localizes the script, passing the experiments from PHP to JS to be read. Let me change it's name to include localize and move the enqueue there.

`acf-experiments` reads that data and applies it to the global acf object.

assets/src/js/experiments/index.ts
gziolo: Can you move types to another file to solve the issue? The rest of the code is contained inside IIFE which prevents variables from being exposed in the global namespace.

includes/admin/views/experiments/experiments.php
cbravobernal: Should we do it with variables and filenames too? Now is the best moment to do it.

includes/admin/views/experiments/experiments.php
cbravobernal: We can be more explicit, leaving it empty was using the current admin url page.

assets/src/js/experiments/types.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export interface ACF {
+	experiments?: ExperimentsObject;
+	[ key: string ]: any;
+}
+
+declare global {
+	interface Window {
+		acf: ACF;
+		acfExperiments?: ExperimentsData; // Variable created by wp_localize_script in admin-experiments.php L150.
```
gziolo: Nit: I don't think exact line reference will age well. It could be also the case with the file name, so if you want to provide more context, you can list the PR instead.

assets/src/js/_acf-experiments.ts
priethor: Actually, you can use the `acf_localize_data` PHP function to pass data directly to `window.afc.data`.

cbravobernal:
Made a huge refactor:
 - Removed al TS and JS involved.
 - Renamed everything to "Beta features"

assets/src/js/_acf-experiments.ts
cbravobernal: Done!

includes/admin/experiments/class-scf-admin-experiment-editor-sidebar.php
cbravobernal: Done.

assets/src/js/experiments/index.ts
cbravobernal: Removed all JS of the experiment :)

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
+<?php // phpcs:disable Universal.Files.SeparateFunctionsFromOO.Mixed
+/**
+ * Admin Beta Features
+ *
+ * This file contains the admin beta features functionality for Secure Custom Fields.
+ *
+ * @package    Secure Custom Fields
+ * @since      6.4.3
```
priethor:
```suggestion
 * @since      6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+
+		/**
+		 * Contains an array of admin beta feature instances.
+		 *
+		 * @var array
+		 */
+		private $beta_features = array();
+
+		/**
+		 * The active beta feature.
```
priethor: What does this mean? Just by looking at the declaration and comment, before checking the rest of the code, it sounds like only one beta feature can be active at a time.

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+		 */
+		public function __construct() {
+			// Temporarily disabled - will be enabled when beta feature is ready
+			// add_action( 'admin_menu', array( $this, 'admin_menu' ), 20 );
+		}
+
+		/**
+		 * This function will store an beta feature class instance in the beta features array.
+		 *
+		 * @since   SCF 6.4.2
```
priethor:
```suggestion
		 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+		 */
+		public function register_beta_feature( $beta_feature ) {
+			$instance                               = new $beta_feature();
+			$this->beta_features[ $instance->name ] = $instance;
+		}
+
+		/**
+		 * This function will return an beta feature class or null if not found.
+		 *
+		 * @since   SCF 6.4.2
```
priethor:
```suggestion
		 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+		 * @return  mixed (SCF_Admin_Beta_Feature|null)
+		 */
+		public function get_beta_feature( $name ) {
+			return isset( $this->beta_features[ $name ] ) ? $this->beta_features[ $name ] : null;
+		}
+
+		/**
+		 * This function will return an array of all beta feature instances.
+		 *
+		 * @since   SCF 6.4.2
```
priethor:
```suggestion
		 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+			// Include beta features
+			$this->include_beta_features();
+
+			return $this->beta_features;
+		}
+
+		/**
+		 * Localizes the beta features data.
+		 *
+		 * @since   SCF 6.4.2
```
priethor:
```suggestion
		 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+		 *
+		 * @return  void
+		 */
+		public function localize_beta_features() {
+			$beta_features = array();
+			foreach ( $this->get_beta_features() as $name => $beta_feature ) {
+				$beta_features[ $name ] = $beta_feature->is_enabled();
+			}
+
+			acf_localize_data(
```
priethor, line 94: We should do a follow-up and create a REST endpoint for this.

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+					'betaFeatures' => $beta_features,
+				)
+			);
+		}
+
+		/**
+		 * This function will add the SCF beta features menu item to the WP admin
+		 *
+		 * @type    action (admin_menu)
+		 * @since   SCF 6.4.2
```
priethor:
```suggestion
		 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+			$page = add_submenu_page( 'edit.php?post_type=acf-field-group', __( 'Beta Features', 'secure-custom-fields' ), __( 'Beta Features', 'secure-custom-fields' ), acf_get_setting( 'capability' ), 'scf-beta-features', array( $this, 'html' ) );
+
+			add_action( 'load-' . $page, array( $this, 'load' ) );
+			add_action( 'admin_enqueue_scripts', array( $this, 'localize_beta_features' ), 20 );
+		}
+
+		/**
+		 * Loads the admin beta features page.
+		 *
+		 * @since   SCF 6.4.2
```
priethor:
```suggestion
		 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+		 *
+		 * @return  void
+		 */
+		public function load() {
+			add_filter( 'admin_body_class', array( $this, 'admin_body_class' ) );
+			// disable filters (default to raw data)
```
priethor: What's the reason to disable the filters?

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+			// Include and register beta features before checking submit
+			$this->include_beta_features();
+
+			$this->check_submit();
+		}
+
+		/**
+		 * Modifies the admin body class.
+		 *
+		 * @since SCF 6.4.2
```
priethor:
```suggestion
		 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+		 */
+		public function admin_body_class( $classes ) {
+			$classes .= ' acf-admin-page';
+			return $classes;
+		}
+
+		/**
+		 * Includes various beta feature-related files.
+		 *
+		 * @since   SCF 6.4.2
```
priethor:
```suggestion
		 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+
+			add_action( 'scf/include_admin_beta_features', array( $this, 'register_beta_features' ) );
+
+			do_action( 'scf/include_admin_beta_features' );
+		}
+
+		/**
+		 * Register default beta features.
+		 *
+		 * @since   SCF 6.4.2
```
priethor:
```suggestion
		 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+		}
+
+		/**
+		 * Includes various beta feature-related files.
+		 *
+		 * @since   SCF 6.4.2
+		 *
+		 * @return  void
+		 */
+		public function include_beta_features() {
```
priethor: Does this function need to be public?

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+		 * @return  void
+		 */
+		public function register_beta_features() {
+			scf_register_admin_beta_feature( 'SCF_Admin_Beta_Feature_Editor_Sidebar' );
+		}
+
+		/**
+		 * Verifies the nonces and submits the value if it passes.
+		 *
+		 * @since   SCF 6.4.2
```
priethor:
```suggestion
		 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+
+			if ( $updated ) {
+				add_action( 'admin_notices', array( $this, 'admin_notices' ) );
+			}
+		}
+
+		/**
+		 * Display admin notices.
+		 *
+		 * @since   SCF 6.4.2
```
priethor:
```suggestion
		 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+			<div class="notice notice-success is-dismissible">
+				<p><?php esc_html_e( 'Beta feature settings updated successfully.', 'secure-custom-fields' ); ?></p>
+			</div>
+			<?php
+		}
+
+		/**
+		 * Admin Beta Features html
+		 *
+		 * @since   SCF 6.4.2
```
priethor:
```suggestion
		 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+				add_meta_box( 'scf-admin-beta-feature-' . $name, acf_esc_html( $beta_feature->title ), array( $this, 'metabox_html' ), $screen->id, 'normal', 'default', array( 'beta_feature' => $name ) );
+			}
+
+			acf_get_view( 'beta-features/beta-features', $view );
+		}
+
+		/**
+		 * Output the metabox HTML for specific beta features
+		 *
+		 * @since SCF 6.4.2
```
priethor:
```suggestion
		 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+
+	// initialize
+	acf()->admin_beta_features = new SCF_Admin_Beta_Features();
+endif; // class_exists check
+
+/**
+ * Alias of acf()->admin_beta_features->register_beta_feature()
+ *
+ * @type    function
+ * @since   SCF 6.4.2
```
priethor:
```suggestion
 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+ */
+function scf_register_admin_beta_feature( $beta_feature ) {
+	acf()->admin_beta_features->register_beta_feature( $beta_feature );
+}
+
+/**
+ * This function will return the admin URL to the beta features page
+ *
+ * @type    function
+ * @since   SCF 6.4.2
```
priethor:
```suggestion
 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+ */
+function scf_get_admin_beta_features_url() {
+	return admin_url( 'edit.php?post_type=acf-field-group&page=scf-beta-features' );
+}
+
+/**
+ * This function will return the admin URL to a specific beta feature page
+ *
+ * @type    function
+ * @since   SCF 6.4.2
```
priethor:
```suggestion
 * @since   SCF 6.5.0
```

includes/admin/beta-features/class-scf-beta-feature-editor-sidebar.php
```diff
@@ -0,0 +1,52 @@
+<?php
+/**
+ * Editor Sidebar Beta Feature
+ *
+ * This beta feature allows moving field group elements to the editor sidebar.
+ *
+ * @package    Secure Custom Fields
+ * @since      6.4.3
```
priethor:
```suggestion
 * @since   SCF 6.5.0
```

includes/admin/beta-features.php
```diff
@@ -0,0 +1,316 @@
...
+		 * Admin Beta Features html
+		 *
+		 * @since   SCF 6.4.2
+		 *
+		 * @return  void
+		 */
+		public function html() {
+			// vars
+			$screen = get_current_screen();
+			$active = acf_maybe_get_GET( 'beta_feature' );
```
priethor: Do we need to save the `$active` value at the class level, considering it's only used inside this function?

priethor (changes requested):
Let's update the `since` versions before merging 😅 

Not a blocker, but could we make more functions private and minimize the amount of public functions exposed as an API?

priethor (changes requested):
Let's update the `since` versions before merging 😅 

Not a blocker, but could we make more functions private and minimize the amount of public functions exposed as an API?

includes/admin/beta-features.php
cbravobernal: No, also, it is just a CSS feature, we can completely remove it. Nice catch.

includes/admin/beta-features.php
cbravobernal: It was initally added to prevent in a future with data that can be filtered by different plugins. It is more a security method, as we won't know how this new features will work with different scf extensions, but for now we can remove it too.

cbravobernal:
> Not a blocker, but could we make more functions private and minimize the amount of public functions exposed as an API?

I made everything as private as possible. But the approach of each experiment having its custom class, extending the main one, needs to be public 🤔 

Maybe we can move that approach to a different one, where every experiment is inside that class. That way we could avoid external devs from creating experiments. Less modular, but less extensible too.

priethor (approved):
> But the approach of each experiment having its custom class, extending the main one, needs to be public 🤔

Yep, and the same goes with the hooks. I was just surprised by the amount of "Since version X" I found 😅 

This is good to merge and start using, in my opinion. While these beta features are not meant for extenders of the plugin, I still think it would be good to have a short readme to explain SCF contributors how to add new beta features. Not a blocker, though!

cbravobernal:
> While these beta features are not meant for extenders of the plugin, I still think it would be good to have a short readme to explain SCF contributors how to add new beta features. Not a blocker, though!

I prefer to start with a couple of experiments to test and iterate and then, when we are 100% sure, add the readme. 😅 

Thee easiest and fastest way to check for an experiment to exist is just to check the option, so we may save some functions.

PR #114, opened 2025-04-23
Allow config errors to be display to users

### LINKED ISSUE #85: Shortcode Bug

pkevan:
Shows errors when user is logged in, and has appropriate capabilities.

Fixes #85.

## Testing instructions

1. Add field to new install of SCF
2. Use shortcode [acf] to output field via [acf field=""]
3. Observe that with an administrator role, no output is display
4. Apply this PR
5. Re-test output, and a display error should be displayed in place of the field output
6. Check while logged out that no output is display as was show when logged in as administrator

Note: considered adding this to (https://github.com/WordPress/secure-custom-fields/blob/a1257ed19fbcf335411f0d0567c90154c78084d9/includes/api/api-template.php#L1053), but given the additional filter, it didn't seem appropriate, although happy to be convinced otherwise.

### DIFF
--- a/includes/api/api-template.php
+++ b/includes/api/api-template.php
@@ -1084,7 +1084,7 @@ function acf_shortcode( $atts ) {
 	}
 
 	if ( isset( $field['allow_in_bindings'] ) && ! $field['allow_in_bindings'] ) {
-		if ( is_preview() || current_user_can( $display_errors_to_user )) {
+		if ( is_preview() || current_user_can( $display_errors_to_user ) ) {
 			return apply_filters( 'acf/shortcode/field_not_allowed_message', '[' . esc_html__( 'The requested ACF field is not allowed to be output in bindings or the ACF Shortcode.', 'secure-custom-fields' ) . ']' );
 		} else {
 			return;
--- a/includes/api/api-template.php
+++ b/includes/api/api-template.php
@@ -1072,8 +1072,11 @@ function acf_shortcode( $atts ) {
 
 	$field_type = is_array( $field ) && isset( $field['type'] ) ? $field['type'] : 'text';
 
+	// Allow config errors to show for appropriate caps checked users.
+	$display_errors_to_user = apply_filters( 'acf/shortcode/display_admin_errors', 'manage_options' );
+
 	if ( ! acf_field_type_supports( $field_type, 'bindings', true ) ) {
-		if ( is_preview() ) {
+		if ( is_preview() || current_user_can( $display_errors_to_user ) ) {
 			return apply_filters( 'acf/shortcode/field_not_supported_message', '[' . esc_html__( 'The requested ACF field type does not support output in bindings or the ACF Shortcode.', 'secure-custom-fields' ) . ']' );
 		} else {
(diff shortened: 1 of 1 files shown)


priethor:
Hi Paul! Can you share brief instructions in the PR description to test the PR?

pkevan:
> Hi Paul! Can you share brief instructions in the PR description to test the PR?

Most of the details are in the original issue: https://github.com/WordPress/secure-custom-fields/issues/85#issue-2934751854 but can adjust the PR if needed a bit later.

priethor:
I would suggest adding, in general, a short section with steps to test it and the expected outcome with and without the PR, similar to Gutenberg PRs. Even though the original ticket wasn't too hard to follow, it still required me to read all the comments to get the context 😅

pkevan:
> I would suggest adding, in general, a short section with steps to test it and the expected outcome with and without the PR, similar to Gutenberg PRs. Even though the original ticket wasn't too hard to follow, it still required me to read all the comments to get the context 😅

Added in test instruction.

priethor:
For documentation purposes, note that I had to do some (obvious) extra steps:
- Enable shortcodes programmatically.
- The Field Group is assigned to posts

With this branch, the error message appears indeed. However, I also see it when using the shortcode to display a Text Field, which, in theory, is supported by the shortcode. Am I missing something?

pkevan:
> However, I also see it when using the shortcode to display a Text Field, which, in theory, is supported by the shortcode. Am I missing something?

🤔 presumably this works as expected without the patch?

pkevan:
> However, I also see it when using the shortcode to display a Text Field, which, in theory, is supported by the shortcode.

In order to not see the message, the field needs to be enabled via the Editor UI radio:

Screenshot 2025-04-30 at 15 23 44

Or were you trying another way?

pkevan:
I wonder how much value there would be in enabling the shortcode via a setting, rather than requiring some code to be added?

priethor (approved):
Sorry for the late reply, my issue was related to the `Allow Access to Value in Editor UI` as I was seeing a different kind of error message:
- With this PR:
  - Disabling the checkbox shows the message `[The requested ACF field is not allowed to be output in bindings or the ACF Shortcode.]` message from the preview, or if the user has enough permissions (not for unlogged users).
  - Enabling the checkbox hides the message.
- Without this PR:
  - The message doesn't show regardless of the checkbox status or the user.

PR #118, opened 2025-04-23
Update initialization of the `acfL10n` object to ensure it's available globally

priethor:
## What and why
When working with translations in the client, I ran into some issues with the `acfL10n` not being initialized properly in my scope. This change ensures `acfL10n` is available in all browsers and conditions.

## Testing instructions
Just use SCF and notice that nothing changed for the current features. However, this change will come in handy in the new features we are working on inside the editor.

### DIFF
--- a/assets/src/js/_acf.js
+++ b/assets/src/js/_acf.js
@@ -673,9 +673,8 @@
 	 *  @return	string Translated text.
 	 */
 
-	if ( window.acfL10n == undefined ) {
-		acfL10n = {};
-	}
+	// Make sure a global acfL10n object exists to prevent errors in other scopes
+	window.acfL10n = window.acfL10n || {};
 
 	acf.__ = function ( text ) {
 		return acfL10n[ text ] || text;


cbravobernal (approved):
LGTM

assets/src/js/_acf.js
```diff
@@ -673,9 +673,8 @@
 	 *  @return	string Translated text.
 	 */
 
-	if ( window.acfL10n == undefined ) {
-		acfL10n = {};
-	}
+	// Make sure a global acfL10n object exists to prevent errors in other scopes
```
pkevan, line 676: Outside the scope of this PR, but we should probably add some checks on coding style outside PHP files, e.g. to create a consistent way for commenting etc (https://github.com/WordPress/secure-custom-fields/pull/118/files#diff-2c39fc5eb9675f3ca4efecbc4a219c3c560c52ad26d42d45217ad878755a59dfL852 - note the ending stop symbol)

priethor: That's a good catch, thanks for noticing 💯

pkevan: It's not the only inconsistency, don't worry!

PR #121, opened 2025-04-25
Add Command Palette support

priethor:
##  What

This PR adds support for WordPress Command Palette/Core Commands (Cmd+K / Ctrl+K interface) by registering:
- Core commands to navigate to SCF admin pages
- "View All" commands for each Custom Post Type registered with SCF
- "Add new" for each Custom Post Type registered with SCF

## Implementation details

The commands are divided into admin and CPT command files.
- Core commands are only registered when the current user has admin permissions
- CPT commands are registered only when these conditions are met:
  - The current user has enough capabilities to edit the CPT
  - The CPT has enabled the `show_ui` setting

To prevent errors and optimize performance, JavaScript files for core and/or CPT commands will only be enqueued when needed.

##  Testing Instructions

  1. Ensure you're using WordPress 6.3+ (Commands API was introduced in 6.3)
  2. Ensure you have registered at least one CPT (e.g., "Books")
  3. Press Cmd+K (Mac) or Ctrl+K (Windows) in the Block Editor
  4. Type "SCF" or "Taxonomies" to see SCF-related admin commands
  5. Type your CPT name ("Books") to see its related commands ("View books", "Add book")
  6. Verify that commands correctly navigate to their respective pages
  7. As an administrator, verify you can see both admin commands and CPT ones
  8. As a user with limited permissions, verify that you only see commands you have access to. For example, an `editor` should only see CPT commands
  9. Test with a WordPress version earlier than 6.3 to confirm that no errors occur

## Recording

https://github.com/user-attachments/assets/fb85b69c-fd45-49e4-9918-2c39cc17f50c

## Followups
- Consider registering commands with `useCommandLoader` instead, which was introduced in WP 6.4.
- Add tests
- Implement a new REST endpoint for CPTs to avoid using `acf_localize_data`
- Migrate to TypeScript

### DIFF
--- a/assets/src/js/commands/admin-commands.js
+++ b/assets/src/js/commands/admin-commands.js
@@ -1,5 +1,5 @@
 /**
- * SCF Admin Commands
+ * Admin Commands
  *
  * Core WordPress commands for Secure Custom Fields administration.
  * This file registers navigation commands for all primary SCF admin screens,
@@ -9,19 +9,7 @@
  */
 
 wp.domReady( () => {
-	// Make sure required WordPress dependencies are available
-	// This ensures we only register commands where the commands API is supported
-	if (
-		! wp.data ||
-		! wp.data.dispatch ||
-		! wp.data.dispatch( 'core/commands' ) ||
-		typeof wp.commands === 'undefined'
-	) {
-		return;
-	}
-
-	// Wait for ACF to be ready
-	if ( typeof acf === 'undefined' ) {
+	if ( ! wp.data?.dispatch?.( 'core/commands' ) || ! acf?.data ) {
 		return;
 	}
 
--- a/assets/src/js/commands/admin-commands.js
+++ b/assets/src/js/commands/admin-commands.js
@@ -5,7 +5,7 @@
  * This file registers navigation commands for all primary SCF admin screens,
  * enabling quick access through the WordPress commands interface (Cmd+K / Ctrl+K).
  *
- * @since 6.5.0
+ * @since SCF 6.5.0
  */
 
 /**
--- a/assets/src/js/commands/admin-commands.js
+++ b/assets/src/js/commands/admin-commands.js
@@ -38,7 +38,7 @@ queue.add( context, () => {
 		{
 			name: 'field-groups',
 			label: __( 'Field Groups', 'secure-custom-fields' ),
-			url: `${ adminUrl }edit.php?post_type=acf-field-group`,
+			url: 'edit.php?post_type=acf-field-group',
 			icon: 'layout',
 			description: __(
 				'SCF: View and manage custom field groups',
@@ -54,7 +54,7 @@ queue.add( context, () => {
 		{
 			name: 'new-field-group',
 			label: __( 'Create New Field Group', 'secure-custom-fields' ),
-			url: `${ adminUrl }post-new.php?post_type=acf-field-group`,
+			url: 'post-new.php?post_type=acf-field-group',
 			icon: 'plus',
 			description: __(
 				'SCF: Create a new field group to organize custom fields',
@@ -71,7 +71,7 @@ queue.add( context, () => {
 		{
 			name: 'post-types',
 			label: __( 'Post Types', 'secure-custom-fields' ),
-			url: `${ adminUrl }edit.php?post_type=acf-post-type`,
+			url: 'edit.php?post_type=acf-post-type',
 			icon: 'admin-post',
 			description: __(
 				'SCF: Manage custom post types',
@@ -82,7 +82,7 @@ queue.add( context, () => {
 		{
 			name: 'new-post-type',
 			label: __( 'Create New Post Type', 'secure-custom-fields' ),
-			url: `${ adminUrl }post-new.php?post_type=acf-post-type`,
+			url: 'post-new.php?post_type=acf-post-type',
 			icon: 'plus',
 			description: __(
 				'SCF: Create a new custom post type',
@@ -93,7 +93,7 @@ queue.add( context, () => {
 		{
 			name: 'taxonomies',
 			label: __( 'Taxonomies', 'secure-custom-fields' ),
-			url: `${ adminUrl }edit.php?post_type=acf-taxonomy`,
+			url: 'edit.php?post_type=acf-taxonomy',
 			icon: 'category',
 			description: __(
 				'SCF: Manage custom taxonomies for organizing content',
@@ -104,7 +104,7 @@ queue.add( context, () => {
 		{
 			name: 'new-taxonomy',
 			label: __( 'Create New Taxonomy', 'secure-custom-fields' ),
-			url: `${ adminUrl }post-new.php?post_type=acf-taxonomy`,
+			url: 'post-new.php?post_type=acf-taxonomy',
 			icon: 'plus',
 			description: __(
 				'SCF: Create a new custom taxonomy',
@@ -122,7 +122,7 @@ queue.add( context, () => {
 		{
 			name: 'options-pages',
 			label: __( 'Options Pages', 'secure-custom-fields' ),
-			url: `${ adminUrl }edit.php?post_type=acf-ui-options-page`,
+			url: 'edit.php?post_type=acf-ui-options-page',
 			icon: 'admin-settings',
 			description: __(
 				'SCF: Manage custom options pages for global settings',
@@ -133,7 +133,7 @@ queue.add( context, () => {
 		{
 			name: 'new-options-page',
 			label: __( 'Create New Options Page', 'secure-custom-fields' ),
-			url: `${ adminUrl }post-new.php?post_type=acf-ui-options-page`,
+			url: 'post-new.php?post_type=acf-ui-options-page',
 			icon: 'plus',
 			description: __(
 				'SCF: Create a new custom options page',
@@ -144,7 +144,7 @@ queue.add( context, () => {
 		{
 			name: 'tools',
 			label: __( 'SCF Tools', 'secure-custom-fields' ),
-			url: `${ adminUrl }admin.php?page=acf-tools`,
+			url: 'admin.php?page=acf-tools',
 			icon: 'admin-tools',
 			description: __(
 				'SCF: Access SCF utility tools',
@@ -155,7 +155,7 @@ queue.add( context, () => {
 		{
 			name: 'import',
 			label: __( 'Import SCF Data', 'secure-custom-fields' ),
-			url: `${ adminUrl }admin.php?page=acf-tools&tool=import`,
+			url: 'admin.php?page=acf-tools&tool=import',
 			icon: 'upload',
 			description: __(
 				'SCF: Import field groups, post types, taxonomies, and options pages',
@@ -166,7 +166,7 @@ queue.add( context, () => {
 		{
 			name: 'export',
 			label: __( 'Export SCF Data', 'secure-custom-fields' ),
-			url: `${ adminUrl }admin.php?page=acf-tools&tool=export`,
+			url: 'admin.php?page=acf-tools&tool=export',
 			icon: 'download',
 			description: __(
 				'SCF: Export field groups, post types, taxonomies, and options pages',
@@ -185,7 +185,7 @@ queue.add( context, () => {
 			description: command.description,
 			keywords: command.keywords,
 			callback: ( { close } ) => {
-				document.location = command.url;
+				document.location = adminUrl + command.url;
 				close();
 			},
 		} );
--- a/assets/src/js/commands/admin-commands.js
+++ b/assets/src/js/commands/admin-commands.js
@@ -15,6 +15,7 @@ import { __ } from '@wordpress/i18n';
 import { createElement } from '@wordpress/element';
 import { Icon } from '@wordpress/components';
 import { dispatch } from '@wordpress/data';
+import { addQueryArgs } from '@wordpress/url';
 
 /**
  * Register admin commands for SCF
@@ -31,7 +32,8 @@ const registerAdminCommands = () => {
 		{
 			name: 'field-groups',
 			label: __( 'Field Groups', 'secure-custom-fields' ),
-			url: 'edit.php?post_type=acf-field-group',
+			url: 'edit.php',
+			urlArgs: { post_type: 'acf-field-group' },
 			icon: 'layout',
 			description: __(
 				'SCF: View and manage custom field groups',
@@ -47,7 +49,8 @@ const registerAdminCommands = () => {
 		{
 			name: 'new-field-group',
 			label: __( 'Create New Field Group', 'secure-custom-fields' ),
-			url: 'post-new.php?post_type=acf-field-group',
+			url: 'post-new.php',
+			urlArgs: { post_type: 'acf-field-group' },
 			icon: 'plus',
 			description: __(
 				'SCF: Create a new field group to organize custom fields',
@@ -64,7 +67,8 @@ const registerAdminCommands = () => {
 		{
 			name: 'post-types',
 			label: __( 'Post Types', 'secure-custom-fields' ),
-			url: 'edit.php?post_type=acf-post-type',
+			url: 'edit.php',
+			urlArgs: { post_type: 'acf-post-type' },
 			icon: 'admin-post',
 			description: __(
 				'SCF: Manage custom post types',
@@ -75,7 +79,8 @@ const registerAdminCommands = () => {
 		{
 			name: 'new-post-type',
 			label: __( 'Create New Post Type', 'secure-custom-fields' ),
-			url: 'post-new.php?post_type=acf-post-type',
+			url: 'post-new.php',
+			urlArgs: { post_type: 'acf-post-type' },
 			icon: 'plus',
 			description: __(
 				'SCF: Create a new custom post type',
@@ -86,7 +91,8 @@ const registerAdminCommands = () => {
 		{
 			name: 'taxonomies',
 			label: __( 'Taxonomies', 'secure-custom-fields' ),
-			url: 'edit.php?post_type=acf-taxonomy',
+			url: 'edit.php',
+			urlArgs: { post_type: 'acf-taxonomy' },
 			icon: 'category',
 			description: __(
 				'SCF: Manage custom taxonomies for organizing content',
@@ -97,7 +103,8 @@ const registerAdminCommands = () => {
 		{
 			name: 'new-taxonomy',
 			label: __( 'Create New Taxonomy', 'secure-custom-fields' ),
-			url: 'post-new.php?post_type=acf-taxonomy',
+			url: 'post-new.php',
+			urlArgs: { post_type: 'acf-taxonomy' },
 			icon: 'plus',
 			description: __(
 				'SCF: Create a new custom taxonomy',
@@ -115,7 +122,8 @@ const registerAdminCommands = () => {
 		{
 			name: 'options-pages',
 			label: __( 'Options Pages', 'secure-custom-fields' ),
-			url: 'edit.php?post_type=acf-ui-options-page',
+			url: 'edit.php',
+			urlArgs: { post_type: 'acf-ui-options-page' },
 			icon: 'admin-settings',
 			description: __(
 				'SCF: Manage custom options pages for global settings',
@@ -126,7 +134,8 @@ const registerAdminCommands = () => {
 		{
 			name: 'new-options-page',
 			label: __( 'Create New Options Page', 'secure-custom-fields' ),
-			url: 'post-new.php?post_type=acf-ui-options-page',
+			url: 'post-new.php',
+			urlArgs: { post_type: 'acf-ui-options-page' },
 			icon: 'plus',
 			description: __(
 				'SCF: Create a new custom options page',
@@ -137,7 +146,8 @@ const registerAdminCommands = () => {
 		{
 			name: 'tools',
 			label: __( 'SCF Tools', 'secure-custom-fields' ),
-			url: 'admin.php?page=acf-tools',
+			url: 'admin.php',
+			urlArgs: { page: 'acf-tools' },
 			icon: 'admin-tools',
 			description: __(
 				'SCF: Access SCF utility tools',
@@ -148,7 +158,8 @@ const registerAdminCommands = () => {
 		{
 			name: 'import',
 			label: __( 'Import SCF Data', 'secure-custom-fields' ),
-			url: 'admin.php?page=acf-tools&tool=import',
+			url: 'admin.php',
+			urlArgs: { page: 'acf-tools', tool: 'import' },
 			icon: 'upload',
 			description: __(
 				'SCF: Import field groups, post types, taxonomies, and options pages',
@@ -159,7 +170,8 @@ const registerAdminCommands = () => {
 		{
 			name: 'export',
 			label: __( 'Export SCF Data', 'secure-custom-fields' ),
-			url: 'admin.php?page=acf-tools&tool=export',
+			url: 'admin.php',
+			urlArgs: { page: 'acf-tools', tool: 'export' },
 			icon: 'download',
 			description: __(
 				'SCF: Export field groups, post types, taxonomies, and options pages',
@@ -178,7 +190,9 @@ const registerAdminCommands = () => {
 			description: command.description,
 			keywords: command.keywords,
 			callback: ( { close } ) => {
-				document.location = adminUrl + command.url;
+				document.location = command.urlArgs
+					? addQueryArgs( adminUrl + command.url, command.urlArgs )
+					: adminUrl + command.url;
 				close();
 			},
 		} );
--- a/assets/src/js/commands/admin-commands.js
+++ b/assets/src/js/commands/admin-commands.js
@@ -8,17 +8,25 @@
  * @since 6.5.0
  */
 
-wp.domReady( () => {
-	if ( ! wp.data?.dispatch?.( 'core/commands' ) || ! acf?.data ) {
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { createElement } from '@wordpress/element';
+import { Icon } from '@wordpress/components';
+import { dispatch } from '@wordpress/data';
+import domReady from '@wordpress/dom-ready';
+
+/**
+ * Register admin commands for SCF
+ */
+domReady( () => {
+	if ( ! dispatch( 'core/commands' ) || ! window.acf?.data ) {
 		return;
 	}
 
-	const { __ } = wp.i18n;
-	const { createElement } = wp.element;
-	const { Icon } = wp.components;
-	const commandStore = wp.data.dispatch( 'core/commands' );
-
-	const adminUrl = acf?.data?.admin_url || '';
+	const commandStore = dispatch( 'core/commands' );
+	const adminUrl = window.acf?.data?.admin_url || '';
 
 	const commands = [
 		{
--- a/assets/src/js/commands/admin-commands.js
+++ b/assets/src/js/commands/admin-commands.js
@@ -15,12 +15,18 @@ import { __ } from '@wordpress/i18n';
 import { createElement } from '@wordpress/element';
 import { Icon } from '@wordpress/components';
 import { dispatch } from '@wordpress/data';
-import domReady from '@wordpress/dom-ready';
+import { createQueue } from '@wordpress/priority-queue';
+
+/**
+ * Initialize deferred execution queue and context
+ */
+const queue = createQueue();
+const context = {};
 
 /**
  * Register admin commands for SCF
  */
-domReady( () => {
+queue.add( context, () => {
 	if ( ! dispatch( 'core/commands' ) || ! window.acf?.data ) {
 		return;
 	}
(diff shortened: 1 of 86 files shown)


assets/src/js/commands/admin-commands.js
```diff
@@ -0,0 +1,179 @@
...
+ * Admin Commands
+ *
+ * Core WordPress commands for Secure Custom Fields administration.
+ * This file registers navigation commands for all primary SCF admin screens,
+ * enabling quick access through the WordPress commands interface (Cmd+K / Ctrl+K).
+ *
+ * @since 6.5.0
+ */
+
+wp.domReady( () => {
```
gziolo: It's possible to avoid using `domReady` and set the strategy for the registered script to `defer`. However, I see that `defer` [got introduced in WordPress 6.3](https://make.wordpress.org/core/2023/07/14/registering-scripts-with-async-and-defer-attributes-in-wordpress-6-3/), but the plugin supports WordPress 6.0 and higher.

assets/src/js/commands/admin-commands.js
```diff
@@ -0,0 +1,179 @@
...
+	if ( ! wp.data?.dispatch?.( 'core/commands' ) || ! acf?.data ) {
+		return;
+	}
+
+	const { __ } = wp.i18n;
+	const { createElement } = wp.element;
+	const { Icon } = wp.components;
+	const commandStore = wp.data.dispatch( 'core/commands' );
```
gziolo: It is possible to use import statements to let the build tackle detecting all dependencies. @cbravobernal added all necessary code to automate it. See:
- https://github.com/WordPress/secure-custom-fields/pull/86

assets/src/js/commands/admin-commands.js
```diff
@@ -0,0 +1,179 @@
...
+	commands.forEach( ( command ) => {
+		commandStore.registerCommand( {
+			name: 'acf/' + command.name,
+			label: command.label,
+			icon: createElement( Icon, { icon: command.icon } ),
+			context: 'admin',
+			description: command.description,
+			keywords: command.keywords,
+			callback: ( { close } ) => {
+				document.location = command.url;
```
gziolo: Nit: the following could help remove some duplication from the command config a bit:
```suggestion
				document.location = adminUrl + command.url;
```

includes/assets.php
```diff
@@ -243,6 +243,30 @@ public function register_scripts() {
 				);
 			}
 
+			// Register core scripts
```
gziolo: There is most likely some code duplication as the same script handles are listed above. @cbravobernal should have more insights here. I also think that we could reuse the asset files generate in the build process together with `commands-admin` and `commands-custom-post-types`. I also belive this script handles need to be prefixed as it is only possible to register one script under a given script handle. These names are too general.

includes/assets.php
```diff
@@ -262,6 +286,13 @@ public function register_scripts() {
 			 * @param   string $suffix The potential ".min" filename suffix.
 			 */
 			do_action( 'acf/register_scripts', $version, $suffix );
+
+			// Ensure WordPress scripts needed for command palette integration are available
+			if ( function_exists( 'wp_enqueue_script' ) ) {
+				wp_enqueue_script( 'wp-commands' );
+				wp_enqueue_script( 'wp-i18n' );
+				wp_enqueue_script( 'wp-dom-ready' );
+			}
```
gziolo: This shouldn't be necessary if they are listed as dependencies for other enqueued scripts.

gziolo (commented):
I left some initial general feedback before trying to better understand how it exactly works. I will try to test it next.

assets/src/js/commands/admin-commands.js
priethor: I can migrate it to use `defer`, since the JavaScript files won't be enqueued in versions earlier than 6.3 anyway, as 6.3 is when the Commands API was also introduced. 👍

gziolo: > Ensure you're using WordPress 6.3+ (Commands API was introduced in 6.3)

So, technically speaking, this feature could be coded with `defer`.

> Consider registering commands with useCommandLoader instead, which was introduced in WP 6.4.

In that case, WP 6.3 users wouldn't be able to use this feature unless you provide 2 different implementations. What's the benefit of using `useCommandLoader`?

assets/src/js/commands/admin-commands.js
```diff
@@ -0,0 +1,179 @@
...
+				'SCF: Export field groups, post types, taxonomies, and options pages',
+				'secure-custom-fields'
+			),
+			keywords: [ 'download', 'json', 'backup', 'migration' ],
+		},
+	];
+
+	commands.forEach( ( command ) => {
+		commandStore.registerCommand( {
+			name: 'acf/' + command.name,
```
gziolo: Nit: What do you think about the following?
```suggestion
			name: 'scf/' + command.name,
```

gziolo:
You can also divide the work into two PRs since it’s fully functional and tests well. Command Palette support doesn’t have to be a Beta Feature because it’s a straightforward enhancement that anyone familiar with the concept should appreciate. 

Follow-up work could include:  
- Refactoring JS files to TS.
- Enhancing asset management to automatically detect dependencies in the build step.  

However, we need to take a closer look at the functionality, particularly the strategy for when these commands become available, depending on the WP version, the presence of custom post types, and user permissions.

assets/src/js/commands/admin-commands.js
priethor: > In that case, WP 6.3 users wouldn't be able to use this feature unless you provide 2 different implementations. What's the benefit of using useCommandLoader?

That's the main reason I didn't go directly with it, it's more straightforward to ensure the commands API is available. I think if we were to use `useCommandLoader`, we should only support that API and not support WP 6.3, having two implementations just for 6.3 seems too specific.

The Command Loader implementation is preferred because it's more performant, as instead of registering the command on page load, it defers the loading until the palette is used and only loads the specific commands that need to be shown to the user.

assets/src/js/commands/admin-commands.js
priethor: Yes, that should be the right prefix. I'll fix it for the CPT commands as well.

assets/src/js/commands/admin-commands.js
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * Core WordPress commands for Secure Custom Fields administration.
+ * This file registers navigation commands for all primary SCF admin screens,
+ * enabling quick access through the WordPress commands interface (Cmd+K / Ctrl+K).
+ *
+ * @since 6.5.0
+ */
+
+wp.domReady( () => {
+	if ( ! wp.data?.dispatch?.( 'core/commands' ) || ! acf?.data ) {
```
gziolo: `@wordpress/commands` depends on `@wordpres/data` so if there is a check on the server whether `wp-commands` is registered then it should be enough as it ensure the rest works as intended. Do I miss some nuance here?

assets/src/js/commands/custom-post-type-commands.js
```diff
@@ -0,0 +1,99 @@
...
+				sprintf(
+					__( 'SCF: View all %s', 'secure-custom-fields' ),
+					pluralLabel
+				),
+			keywords: [
+				'post type',
+				'content',
+				'cpt',
+				postType.name,
+				postType.label || '',
```
gziolo: Passing an empty string as a keyword should be avoided. Would it be possible to add it conditionally?

includes/admin/admin-commands.php
```diff
@@ -0,0 +1,88 @@
...
+
+	acf_localize_data(
+		array(
+			'customPostTypes' => $custom_post_types,
+		)
+	);
+
+	if ( ! empty( $custom_post_types ) ) {
+		wp_enqueue_script( 'commands-custom-post-types' );
+	}
```
gziolo: Nit, it feels like custom post types should be only passed when they are defined.
```suggestion
	if ( ! empty( $custom_post_types ) ) {
		acf_localize_data(
			array(
				'customPostTypes' => $custom_post_types,
			)
		);
		wp_enqueue_script( 'commands-custom-post-types' );
	}
```

priethor: It's not a big difference (empty vs undefined), especially since the JS that uses this data will not be even enqueued, so let's do it as you suggest.

assets/src/js/commands/admin-commands.js
priethor: Yep, this is specifically the `ACF` data. They are both defensive checks in case something didn't go well, but in theory, the commands API and `acf.data` should be available if the file is being executed.

includes/assets.php
priethor: Fixed!

assets/src/js/commands/custom-post-type-commands.js
priethor: Fixed!

assets/src/js/commands/admin-commands.js
priethor: In 887f2cf, I address this and optimize the scripts by:
- Deferring the load with the `defer` strategy.
- Deferring the execution by using the WP priority queue.

includes/assets.php
priethor: Thanks for catching that, they are indeed a leftover after tests.

I've also prefixed the assets again with `scf-` to avoid potential collisions with other plugins.

assets/src/js/commands/admin-commands.js
```diff
@@ -0,0 +1,193 @@
...
+/**
+ * Initialize deferred execution queue and context
+ */
+const queue = createQueue();
+const context = {};
+
+/**
+ * Register admin commands for SCF
+ */
+queue.add( context, () => {
```
youknowriad: It's not really clear to me why you're using "priority-queue" here? Did you actually notice performance issues or something. Also potentially you can just use `requestIdleCallback` directly instead of using a queue with a single callback.

assets/src/js/commands/admin-commands.js
```diff
@@ -0,0 +1,193 @@
...
+	if ( ! dispatch( 'core/commands' ) || ! window.acf?.data ) {
+		return;
+	}
+
+	const commandStore = dispatch( 'core/commands' );
+	const adminUrl = window.acf?.data?.admin_url || '';
```
youknowriad, line 29: which script defines `window.acf` ?

includes/admin/admin-commands.php
```diff
@@ -0,0 +1,87 @@
...
+ * @since 6.5.0
+ */
+function acf_commands_init() {
+	// Only load on admin screens.
+	if ( ! is_admin() ) {
+		return;
+	}
+
+	// Ensure we only load our commands where the WordPress commands API is available.
+	if ( ! wp_script_is( 'wp-commands', 'registered' ) ) {
```
youknowriad, line 30: is this really necessary?

includes/admin/admin-commands.php
```diff
@@ -0,0 +1,87 @@
...
+ * 3. Core administrative commands are only shown to users with SCF admin capabilities.
+ * 4. Custom post type commands are conditionally shown based on edit_posts capability
+ *    for each specific post type.
+ * 5. Post types must have UI enabled (show_ui setting) to appear in commands.
+ *
+ * @since 6.5.0
+ */
+function acf_commands_init() {
+	// Only load on admin screens.
+	if ( ! is_admin() ) {
```
youknowriad: This doesn't seem necessary since you're using `admin_queue_script`

includes/admin/admin-commands.php
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	// Ensure we only load our commands where the WordPress commands API is available.
+	if ( ! wp_script_is( 'wp-commands', 'registered' ) ) {
+		return;
+	}
+
+	$custom_post_types = array();
+
+	if ( function_exists( 'acf_get_acf_post_types' ) ) {
```
youknowriad: Why this check, can't we assume it exists since we're in the SCF codebase?

includes/admin/admin-commands.php
```diff
@@ -0,0 +1,87 @@
...
+					'label'          => $plural_label,
+					'singular_label' => $singular_label,
+					'icon'           => $post_type['menu_icon'] ?? '',
+				);
+			}
+		}
+	}
+
+	if ( ! empty( $custom_post_types ) ) {
+		acf_localize_data(
```
youknowriad, line 66: Personally, I'd argue that we should move away from localized variables like that as much as possible and instead use REST API endpoints. Do SCF has endpoints to retrieve these things, can't we just rely on Core endpoints for post types...?

includes/assets.php
```diff
@@ -243,6 +243,28 @@ public function register_scripts() {
 				);
 			}
 
+			wp_register_script(
+				'scf-commands-admin',
+				acf_get_url( 'assets/build/js/commands/scf-admin' . $suffix . '.js' ),
+				array( 'acf', 'wp-plugins', 'wp-element', 'wp-components', 'wp-data', 'wp-commands', 'wp-i18n', 'wp-dom-ready' ),
```
youknowriad, line 249: why do these scripts depend on the "act" script?

youknowriad (commented):
In parallel to this PR, I wonder if we should implement [this PR](https://github.com/WordPress/gutenberg/pull/54515) within SCF (before it gets into Core) to enable the command palette everywhere.

assets/src/js/commands/admin-commands.js
cbravobernal: The object is defined here:
https://github.com/WordPress/secure-custom-fields/blob/8b71adbd32aa0970c2c693a7c5e150ae2942289b/assets/src/js/_acf.js#L15

And the script is loaded here:
https://github.com/WordPress/secure-custom-fields/blob/8b71adbd32aa0970c2c693a7c5e150ae2942289b/includes/assets.php#L152

https://github.com/WordPress/secure-custom-fields/blob/8b71adbd32aa0970c2c693a7c5e150ae2942289b/includes/assets.php#L237

assets/src/js/commands/admin-commands.js
priethor: I did not notice any performance difference; I implemented the priority queue in parallel to defer the loading [here](https://github.com/WordPress/secure-custom-fields/pull/121#discussion_r2060323077).

includes/admin/admin-commands.php
priethor: That's right, I removed it ✅

includes/admin/admin-commands.php
priethor: This is necessary to avoid trying to access the command palette functions in WP versions earlier than 6.3 or without Gutenberg, in which cases my PR should do nothing or would result in errors.

includes/admin/admin-commands.php
cbravobernal: I think the only endpoints registered are the ones that fetch custom post types and post fields, in every field you can chose to include it or not in the WP default API endpoints for posts fetching, but there are no internal endpoints to retrieve all post types or all post fields created with the plugin.

They are instead localizing the data: 

https://github.com/WordPress/secure-custom-fields/blob/8b71adbd32aa0970c2c693a7c5e150ae2942289b/includes/assets.php#L603

https://github.com/WordPress/secure-custom-fields/blob/8b71adbd32aa0970c2c693a7c5e150ae2942289b/includes/admin/post-types/admin-field-group.php#L151

Moving everything to a REST API could be done, but this needs to define a new API structure, with its own documentation and decide the approach for the store handling ( using Redux based like wp-data, or move to React Query or something similar )

What's your opinion on that @youknowriad ?

includes/assets.php
cbravobernal: I guess because they need the acf object to be there, as its being extended with the localized data.

includes/admin/admin-commands.php
priethor: There are two issues why using the Rest API would require some extra work and I didn't consider for this:
- While it does a PHP function to retrieve SCF Post types (`acf_get_acf_post_types`), it doesn't have a custom enpoint to do so, just WordPress` built in endpoints.
- More importantly, in order to show CPTs in the REST API, each CPT needs to enable the `Show in REST API` setting.

Therefore, my guess is we will need custom SCF endpoints for the admin that show all, as we will encounter a similar issue when implementing DataViews. What do you think about this approach, Riad?

includes/assets.php
priethor: To maintain the proper script loading order and ensure things like `window.acf.data` already exist.

assets/src/js/commands/admin-commands.js
youknowriad: Let's just get rid of it. I think `defer` is enough.

includes/admin/admin-commands.php
youknowriad: ok I didn't realize we support old versions like that in scf

includes/admin/admin-commands.php
youknowriad: We're on WordPress, so for me, we should be use the entities and core-data here. no need to duplicate things that exist. 

> More importantly, in order to show CPTs in the REST API, each CPT needs to enable the Show in REST API setting.

For me this is something that we should try to get rid of. Every CPT should be visible in the endpoints, I think this is a setting of another age. 

That said, for now, a custom endpoint/entity could work if we're not ready to make all CPTs created by SCF show_in_rest true. Are there any valid reasons for not doing it?

includes/admin/admin-commands.php
priethor: There's an option to disable post type entirely, but checking the codebase confirms the function is loaded unconditionally (just returning empty if post types are disabled), so it is indeed safe to assume it always exists. 👍

includes/assets.php
youknowriad: I don't think we should be loading `acf` scripts in pages that might not need it. And the command palette itself can be rendered anywhere and ACF might not be loaded there. 

Do we need rely on `acf` just for the localization part? In other words, if we relied on an endpoint, we would need this right?

priethor: Yep, if we relied on an endpoint we wouldn't need this. The only "real" dependency on ACF is to get the CPTs created by SCF from the DB.

includes/admin/admin-commands.php
priethor: > Are there any valid reasons for not doing it?

If by this you mean a custom endpoint, the only reason not to do it right now is the extra complexity it adds to this PR, we should probably tackle that separately.

As per the existing setting, I wouldn't change an existing setting as we don't know all the cases or real users (e.g., having CPTs only for the admin to manage private data), though.

assets/src/js/commands/admin-commands.js
priethor: I simplified to a `requestIdleCallback` here before reading your answer. Should I get rid of that, too? It still seems overkill.

includes/admin/admin-commands.php
youknowriad: > As per the existing setting, I wouldn't change an existing setting as we don't know all the cases or real users (e.g., having CPTs only for the admin to manage private data), though.

This is different than `show_in_rest`, this should be mostly handled by the `public` boolean.

youknowriad:
So for me, the PR is fine. But I think we should ask ourselves what we want to do with these post types and any SCF UI going forward. Do we want to decouple the server and client and have a consistent data layer to communicate between both (ideally on top of core-data) or do we want to continue with these localized scripts which for me feels like hacks.

An investment in the core-data layer will help this PR but also any future PR that would like to rely on SCF entities in its client UI.

includes/admin/admin-commands.php
priethor: Thinking a bit more about this:
- Right now, we would need to use the `types` endpoint.
- `show_in_rest` determines whether a custom post type gets its own endpoint, but also hides the CPT definition from the `types` endpoint.
- In any case, the `types` endpoint doesn't return any information to know whether a type is a CPT and/or created by SCF (e.g., no `post-type` field)

Therefore, my inclination is to create a new endpoint for custom post types that ignores the `show_in_rest`. This will be needed, too, when migrating the Post Types screen to DataViews.

priethor:
> Do we want to decouple the server and client and have a consistent data layer to communicate between both (ideally on top of core-data) or do we want to continue with these localized scripts which for me feels like hacks.

I don't know if a hack, but it did feel like a dated approach that doesn't align with the current direction of core and modern architecture. I do agree with investing in decoupling server and client; the amount of data required in this PR is minimal, but as we move forward with new functionalities, it will pay off or even become necessary.

How do you all feel about this PR with the current approach, and then follow up with the new endpoint implementation?

includes/admin/admin-commands.php
youknowriad: The ideal scenario is to actually use the `types` endpoint and add "metadata" to the post type. That said, I think it's a good interim solution to have a custom endpoint.

youknowriad:
> How do you all feel about this PR with the current approach, and then follow up with the new endpoint implementation?

I'm ok with this personally, but I would try to avoid the `acf` script dependency, it seems this is not needed at all and we can just use any random global variable.

includes/admin/admin-commands.php
```diff
@@ -0,0 +1,80 @@
...
+	$scf_post_types = acf_get_acf_post_types();
+
+	foreach ( $scf_post_types as $post_type ) {
+		// Skip if post type name is not set (defensive) or post type is inactive.
+		if ( empty( $post_type['post_type'] ) || ( isset( $post_type['active'] ) && ! $post_type['active'] ) ) {
+			continue;
+		}
+
+		$plural_label   = $post_type['labels']['name'] ?? $post_type['label'] ?? $post_type['post_type'];
+		$singular_label = $post_type['labels']['singular_name'] ?? $post_type['singular_label'] ?? $plural_label;
```
mcsf: Is `get_post_type_labels` not enough? If so, can we document why?

includes/admin/admin-commands.php
```diff
@@ -0,0 +1,80 @@
...
+
+		$post_type_obj = get_post_type_object( $post_type['post_type'] );
```
mcsf, line 44: I ask not knowing `acf_get_acf_post_types`: is it ever possible to have a key returned by `acf_get_acf_post_types` whose type object cannot be found? When?

assets/src/js/commands/custom-post-type-commands.js
```diff
@@ -0,0 +1,117 @@
...
+	const postTypes = window.acf.data.customPostTypes;
+
+	postTypes.forEach( ( postType ) => {
+		// Skip invalid post types
+		if ( ! postType?.name ) {
+			return;
+		}
+
+		const pluralLabel = postType.label || postType.name;
+		const singularLabel = postType.singular_label || pluralLabel;
```
mcsf: Both the producer and the consumer of the `window.acf.data.customPostTypes` message are implementing their own fallback logic for computing these labels. Can we make it so that only one side (presumably, the producer) does that?

assets/src/js/commands/custom-post-type-commands.js
```diff
@@ -0,0 +1,117 @@
...
+		// Register "View All" command for this post type
+		commandStore.registerCommand( {
+			name: `scf/cpt-${ postType.name }`,
+			label: pluralLabel,
+			icon: createElement( Icon, { icon: 'admin-page' } ),
+			context: 'admin',
+			description:
+				/* translators: %s: Post type plural label */
+				sprintf(
+					__( 'SCF: View all %s', 'secure-custom-fields' ),
```
mcsf: This is going to be a problem in i18n (stitching together fragments of a sentence from variables). Picture this sentence in Castilian and you'll see why: _Ver todos los_ (masc.) + _imágenes_ (fem.)

The answer in WordPress has generally been [expanding label support](https://developer.wordpress.org/reference/functions/get_post_type_labels/) for post types. The most adequate core label for your use case would be `view_items`, but I have no idea how widely it is implemented in SCF tradition, if at all, and its default value is _View Posts_, which will be confusing in this context.

You could argue that this is something for translators to solve for each locale, but I always worry when I see a string like `View all %s` that translators will either make mistakes or be left with an unreasonable burden to make it work in their locale.

How much worse would the UX be with these more i18n-friendly strings?
* `SCF: %s: View all entries` (more natural, IMO)
* `SCF: View all entries of type "%s"` (more robotic, IMO)
* same, but with `s/entries/items/`

Don't let this block you, btw, but give it a thought.

assets/src/js/commands/custom-post-type-commands.js
```diff
@@ -0,0 +1,117 @@
...
+					pluralLabel
+				),
+			keywords: [
+				'post type',
+				'content',
+				'cpt',
+				postType.name,
+				...( postType.label ? [ postType.label ] : [] ),
+			],
```
mcsf:
```suggestion
			keywords: [
				'post type',
				'content',
				'cpt',
				postType.name,
				postType.label,
			].filter( Boolean ),
```

assets/src/js/commands/custom-post-type-commands.js
```diff
@@ -0,0 +1,117 @@
...
+			},
+		} );
+
+		// Register "Add New" command for this post type
+		commandStore.registerCommand( {
+			name: `scf/new-${ postType.name }`,
+			label:
+				/* translators: %s: Post type singular label */
+				sprintf(
+					__( 'Add New %s', 'secure-custom-fields' ),
```
mcsf: See my other comment about translating post labels. Here, it's `add_new` or `add_new_item`.

assets/src/js/commands/custom-post-type-commands.js
```diff
@@ -0,0 +1,117 @@
...
+					singularLabel
+				),
+			icon: createElement( Icon, { icon: 'plus' } ),
+			context: 'admin',
+			description:
+				/* translators: %s: Post type singular label */
+				sprintf(
+					__( 'SCF: Create a new %s', 'secure-custom-fields' ),
```
mcsf: Same.

assets/src/js/commands/custom-post-type-commands.js
```diff
@@ -0,0 +1,117 @@
...
+					singularLabel
+				),
+			keywords: [
+				'add',
+				'new',
+				'create',
+				'content',
+				postType.name,
+				...( postType.label ? [ postType.label ] : [] ),
```
mcsf, line 80: Same suggestion with `[].filter(Boolean)` on the keywords array

mcsf:
> So for me, the PR is fine. But I think we should ask ourselves what we want to do with these post types and any SCF UI going forward. Do we want to decouple the server and client and have a consistent data layer to communicate between both (ideally on top of core-data) or do we want to continue with these localized scripts which for me feels like hacks.

It's better in the long run to have "proper" communication, but — when we get to that — let's ensure that these payloads are preloaded, which is something we shouldn't take for granted in the old-school localize_script method.

priethor:
Thanks, I'll address the feedback and merge the PR; I just submitted [the required changes to the `types` endpoint](https://github.com/WordPress/secure-custom-fields/pull/128) to migrate the Command Palette from using `acf_localiza_data` to fetching the types from the REST API.

includes/admin/admin-commands.php
priethor: In theory it could in very edge cases, but we already checked `empty( $post_type['post_type'] )` [here](https://github.com/WordPress/secure-custom-fields/pull/121/files/71342ac399ed7169e8069cab83b248ea99a9335c#diff-4c2e11a2899b7dd644477d9386b324e9a81fb12be5db7bbee269a2a6338aa145R40), in which case this line wouldn't execute. The later checking of `$post_type_obj` also ensures this line executes properly.

assets/src/js/commands/custom-post-type-commands.js
priethor: I've tried a few approaches, and none seem ideal. I've proceeded with removing the prefixes and making the commands aligned with the wp-admin menu labels:
- Add command: It's now `Add %s`, which is also easier to translate.
- View all command: it's now `View all %s`. Still not ideal to translate, but it's the exact same string (and issue) as the admin menu label.

includes/admin/admin-commands.php
priethor: Great catch, I've now leveraged that core function 👍

assets/src/js/commands/custom-post-type-commands.js
mcsf: > Add command: It's now Add %s, which is also easier to translate.

Sure, in Castilian. ;) In more inflected languages, nouns may reflect their role in the sentence (e.g. nominative case for the subject, accusative for the direct object). In Esperanto:
- Page (nominative): _Paĝo_
- Add new page (accusative): _Aldoni novan paĝon_

> but it's the exact same string (and issue) as the admin menu label.

Not sure what you mean: are you saying it's the same string as the one shown in WP-Admin when you navigate the sidebar, e.g. _Posts -> All Posts_? If so, well then why not use the [actual label](https://core.trac.wordpress.org/browser/trunk/src/wp-includes/post.php?rev=60236#L2049)?

mcsf (approved):
We need to address the i18n issues right after this, but ship it 🚢

priethor:
> We need to address the i18n issues right after this, but ship it 🚢

Thanks for the feedback, based on our last chat, the last commit directly uses the `all_items` and `add_new_items` provided instead, so we don't add any translatable text and reuse the same labels as wp-admin. These already have fallbacks in case they are empty, too.

assets/src/js/commands/custom-post-type-commands.js
priethor: Resolved [here](https://github.com/WordPress/secure-custom-fields/pull/121#issuecomment-2880698229)

PR #124, opened 2025-04-30
Beta feature: Add endpoint to retrieve custom fields of a post type.

cbravobernal:
##  What
Adds a new field `scf_field_groups` to the WordPress REST API's type(s) endpoints, providing information about field groups and fields associated with each post type. Includes metadata such as field group titles, field labels, and field types

 ##  Why
Serves as the foundation to integrate the post metadata in the editor sidebar, as it requires knowing what fields a post type has.

 ## Testing instructions
 1. Ensure REST API is enabled in WordPress
 2. Make a GET request to `/wp-json/wp/v2/types`
 3. Verify each post type has a `scf_field_groups` property containing an array of field groups
 4. For post types with assigned field groups, confirm each field group includes the `title` and `fields`
 5. Test with a post type that has multiple field groups to ensure all are returned
 6. Test with post types that have no field groups to ensure an empty array is returned


includes/rest-api.php
```diff
@@ -1,4 +1,10 @@
 <?php
+/**
+ * REST API
+ *
+ * @package    Secure Custom Fields
+ * @since      6.4.0
```
priethor, line 6: Important note here: using 6.4.0 as we are not introducing the REST API in this PR, only fixing the missing file annotation that resulted in errors.

No action required!

priethor (approved):
Looks good to me!

I'm also working on adding a new parameter to the same endpoint and needed to create the same `class-acf-rest-types-endpoint, so I started adding it in this same PR to avoid collisions but I'll add it in a follow-up in the end.

In the process, I removed the dependency with the beta features as the Command Palette will depend on the types endpoint changes. If we want to hide this behind the beta features, it should only be the part that registers the new fields, not the whole class. What do you think, @cbravobernal ?

cbravobernal:
> What do you think, @cbravobernal ?

Sure, let's only hide the field registration.

Done in https://github.com/WordPress/secure-custom-fields/pull/124/commits/6c7beef9fc8b89326e17625ca38ff23d9e589472

cbravobernal:
I cannot approve my own PR, but looks good.

PR #127, opened 2025-05-08
Fix: Add compatibility with Woo HPOS. Allow order fields.

cbravobernal:
## What

With the last release of WooCommerce, a [HPOS was introduced.](https://woocommerce.com/document/high-performance-order-storage/).

SCF needs some updates to be compatible with these new meta data management. One first bug that appeared is that if any user adds Order fields, these are not shown or updated in the order page. This PR fixes that.

## How to test

You need a Woo (v 9.8.4) installed, at least one product and an order.

- Create a Field, set the location as a Woo Order.

Screenshot 2025-05-10 at 14 43 27

- Check that the field appears, can be filled and updated.
Screenshot 2025-05-10 at 14 44 15

- Check that field can be called in a different PHP function.

## How to test the meta optimization.

These checks are already included in this PR as e2e tests. But feel free to test them manually.

- Create a Field, set the location as comment, option, user and term.
- Check that field appears in any location, check that can be filled, edited and rendered with their respective PHP function.


includes/acf-meta-functions.php
```diff
@@ -4,7 +4,7 @@
  * Returns an array of "ACF only" meta for the given post_id.
  *
  * @date    9/10/18
- * @since   ACF 5.8.0
```
kraftbj: These references were added intentionally since there wasn't a 5.8.0 of SCF.

cbravobernal: Fixed in https://github.com/WordPress/secure-custom-fields/pull/127/commits/827d2d0ecbfe08f5616765353d97c114f952d58c

kraftbj:
Could you add a few testing instructions? I'd like to be sure I'm understanding what you're seeing and aiming to fix.

cbravobernal:
> Could you add a few testing instructions? I'd like to be sure I'm understanding what you're seeing and aiming to fix.

I updated the description to include a small testing instructions. Also, I will finish today some e2e tests to include all test cases.

cbravobernal:
I will make another review after some rest time. The PR is ready, but still needs a double check to remove maybe not needed "AI" generated stuff on e2e tests, any wrong "since", or any duplicated or wrong comment due to duplicating files for tests.

cbravobernal:
I forgot to push the last e2e. Now is ready.

includes/acf-meta-functions.php
priethor: Related to this, I [opened a PR](https://github.com/WordPress/secure-custom-fields/pull/129) to add `SCF` to the annotations

priethor: Related to this, I opened a PR to [add SCF to the annotations](https://github.com/WordPress/secure-custom-fields/pull/129).

includes/meta/comment.php
```diff
@@ -0,0 +1,23 @@
+<?php
```
cbravobernal: This files are duplicated. Will fix it.

includes/Meta/Comment.php
```diff
@@ -0,0 +1,23 @@
+<?php
+/**
+ * Adds support for saving/retrieving values from comment meta.
+ *
+ * @package    SCF
+ * @since      6.5
```
priethor: Can you adapt these annotations to say `SCF 6.5`, to comply with the changes made in #129 ?

secure-custom-fields.php
```diff
@@ -741,6 +752,19 @@ public function acf_plugin_activated() {
...
+
+		/**
+		 * Initializes the ACF WooCommerce HPOS integration.
+		 *
+		 * @since 6.5
+		 *
+		 * @return void
+		 */
+		public function init_hpos_integration() {
+			acf_include( 'includes/forms/form-wc-order.php' );
```
priethor: Should we check if the `custom_order_tables` feature is enabled in WooCommerce before including these?

priethor (approved):

priethor (changes requested):
Since one of the latest commits, class initialization fails: the classes are trying to be loaded from `includes/SCF/` folder, because of the namespace and autoloader configuration, instead of `includes/Meta`.

priethor (changes requested):
Since one of the latest commits, class initialization fails: the classes are trying to be loaded from `includes/SCF/` folder, because of the namespace and autoloader configuration, instead of `includes/Meta`.

cbravobernal:
> Since one of the latest comments, class initialization fails: the classes are trying to be loaded from `includes/SCF/` folder, because of the namespace and autoloader configuration, instead of `includes/Meta`.

Did you do a `composer dump-autoload` on your local system?

includes/Meta/Comment.php
cbravobernal: I though in that case, as the package was SCF already, we didn't need to specify just one line after. But happy to update that.

priethor:
> Did you do a composer dump-autoload on your local system?

I hadn't before, but this didn't help, even though the autoloader setting seems correct.

cbravobernal:
> > Did you do a composer dump-autoload on your local system?
> 
> I hadn't before, but this didn't help, even though the autoloader setting seems correct.

You are right, I removed the SCF Meta on the composer file 🤦‍♂️

secure-custom-fields.php
cbravobernal: The functions already have a backup method if the feature is not enabled.

includes/Meta/Comment.php
cbravobernal: Done!

includes/meta/comment.php
cbravobernal: Fixed.

priethor (approved):
Everything works now as expected, @cbravobernal and I have been testing it for the last couple of hours.

PR #128, opened 2025-05-12
REST API: Add `source` parameter for types endpoint

priethor:
##  What

  - Added a new source parameter to the `/wp/v2/types` REST API endpoint that allows filtering post types by their origin: `core` (WordPress built-in), `scf` (for SCF managed types), or `other` for the rest of CPTs.
  - Works with both collection (`/wp/v2/types`) and single item (`/wp/v2/types/{type}`) endpoints
  - Properly documented in the OpenAPI schema

 ## Why

When working on the Command Palette Integration and considering using the REST API to define CPT-related commands, requesting specifically for the CPTs managed by SCF became a need, and it will be needed, too, when implementing DataViews.

In fact, this branch is directly based on #124 and started in parallel, as we both realized we needed to tweak the Types endpoint if we wanted to leverage the API.

 ## How
- If the `source` parameter is not defined, nothing changes.
- For specific, single-type requests:
  - Hooks to `rest_request_before_callbacks` for the single request, returning 404 if the requested type doesn't meet the criteria.
- For collection requests:
  - Hooks to `rest_prepare_post_type` to filter each post type by source _before_ WordPress prepares it.
  - Hooks to `rest_pre_echo_response` to remove `null` values returned by `rest_prepare_post_type` in the response array.
  - Since the filter `rest_prepare_post_type` fires once per post type, the Post Types belonging to the requested source are calculated in advance in `rest_request_before_callbacks` and reused throughout the whole request.

 ## Testing Instructions

  1. API Endpoint Testing:
    - Make a GET request to `/wp/v2/types` to see all post types
    - Add `?source=core` to filter for WordPress built-in post types only
    - Add `?source=scf` to see only post types managed by SCF
    - Try `?source=other` to see post types registered from other sources
    - Test single post type endpoints like`/wp/v2/types/post?source=core`
  2. Validation Testing:
    - Try an invalid source value (e.g., `?source=invalid`) to verify validation works
    - Confirm a 404 error when requesting a post type with the wrong source (e.g., /wp/v2/types/post?source=scf`
  3. Unit and E2E Tests (_heavily assisted by AI_):
    - Run PHPUnit tests: `composer test:php -- --filter=Test_REST_Types_Endpoint`
    - Run E2E tests: `npm run test:e2e tests/e2e/rest-api-types-endpoint.spec.ts`

## Potential follow-ups
- Add a hook so that other plugins can register their own sources.


includes/rest-api.php
```diff
@@ -3,7 +3,7 @@
  * REST API
  *
  * @package    Secure Custom Fields
- * @since      ACF 6.4.0
+ * @since      6.5.0
```
cbravobernal: Should we specify SCF 6.5.0 . So LLMs don't mess.

tests/e2e/rest-api-types-endpoint.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+      
+      const otherTypes = await requestUtils.rest({
+        path: '/wp/v2/types',
+        params: { source: 'other' }
+      });
+      
+      console.log('All types:', Object.keys(allTypes));
+      console.log('SCF types:', Object.keys(scfTypes));
+      console.log('Core types:', Object.keys(coreTypes));
+      console.log('Other types:', Object.keys(otherTypes));
```
cbravobernal: Let's remove the logs here.

tests/e2e/rest-api-types-endpoint.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+          params: { source: 'scf' }
+        });
+        
+        // Should succeed if our test post type is properly registered with SCF
+        expect(typeWithScfSource).toHaveProperty('slug', customTestType);
+      } else {
+        console.log(`Test post type ${customTestType} not found - skipping SCF single post type tests`);
+      }
+    } catch (error) {
+      console.log(`Error checking SCF test type: ${error.message}`);
```
cbravobernal: We are mixing console.error and console.logs in errors :-)

tests/e2e/rest-api-types-endpoint.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+        path: '/wp/v2/types',
+        params: { source: 'core' }
+      });
+      console.log('Core post types:', Object.keys(coreTypes));
+      
+      const otherTypes = await requestUtils.rest({
+        path: '/wp/v2/types',
+        params: { source: 'other' }
+      });
+      console.log('Other post types:', Object.keys(otherTypes));
```
cbravobernal: Let's remove this logs too.

tests/e2e/rest-api-types-endpoint.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+        expect(inScf || inOther).toBe(true);
+        
+        if (inScf) {
+          console.log(`✅ Perfect! Test post type is in SCF source as desired`);
+        } else if (inOther) {
+          console.log(`⚠️ Acceptable: Test post type in other source (our filter hook didn't work)`);
+        }
+      } else {
+        console.log(`⚠️ Test post type ${customTestType} not found - skipping this part of the test`);
+      }
```
cbravobernal: We don't need this too.

cbravobernal (changes requested):
In my opinion, most of console logs are not necessary. Let's avoid to fill the logs of CI with them. The rest of the PR is working as expected, nice job!! 🚀

includes/rest-api.php
priethor: Good catch, that's the result of a bad merge from trunk 👍

includes/rest-api/class-acf-rest-types-endpoint.php
```diff
@@ -14,19 +14,174 @@
...
+				$result = array_diff(
+					array_keys( get_post_types( array(), 'objects' ) ),
+					array_merge( $core_types, $scf_types )
+				);
+				break;
+			default:
+				$result = array();
+		}
+
+		return $result;
```
cbravobernal, line 156: NIT: We can return on each case.

tests/e2e/plugins/scf-test-setup-post-types.php
```diff
@@ -0,0 +1,108 @@
...
+		),
+	);
+
+	// Create the post type entry in the database using SCF's internal API
+	$result = $instance->update_post( $post_type_config );
+
+	if ( is_array( $result ) && isset( $result['ID'] ) ) {
+		// Mark as created so we don't duplicate it
+		update_option( 'scf_test_post_type_created', true );
+	}
```
cbravobernal: NIT: Delete option on plugin deactivation / uninstall.

cbravobernal (approved):
Nice work!

tests/e2e/plugins/scf-test-setup-post-types.php
priethor: This is actually done [here](https://github.com/WordPress/secure-custom-fields/pull/128/files#diff-cf535a3472864f7bb220c2170bce246251c88a13d79ea1e7fb3b15c0dc4b116fR394), although there was a typo.

What was missing was deleting the post type itself, which I fixed in the latest commit. Since post deletion is done on plugin deactivation on the PHP side, I've moved deleting the option to [the same deactivation hook](https://github.com/WordPress/secure-custom-fields/pull/128/commits/e5588b7cdbc16c11cf95052251508d27bcad72f5#diff-da74914f5a44e4f02443fa49346343c30fc490b7528c53d5599e341450edc563R109) for consistency.

includes/rest-api/class-acf-rest-types-endpoint.php
```diff
@@ -14,19 +14,174 @@
...
+		// Only proceed if source parameter is provided and valid
+		if ( ! $source || ! in_array( $source, array( 'core', 'scf', 'other' ), true ) ) {
+			return $response;
+		}
+
+		// Get post types, calculating once and reusing for the entire request
+		if ( null === $this->cached_post_types ) {
+			$this->cached_post_types = $this->get_source_post_types( $source );
+		}
+		$source_post_types = $this->cached_post_types;
```
mcsf: In practice, the way the REST API works is that one request corresponds to one fresh execution of the WP stack, including running through `SCF_Rest_Types_Endpoint`'s constructor and down that stack. However, semantically, `cached_post_types` is an instance property of a class that represents an endpoint, not a request. Here's my concern when I read this chunk:

Is there any scenario (e.g. request batching or sequential `rest_do_request` calls) where it's possible to process more than one request with the same endpoint instance? If so, this cache will need invalidation. At the very least, it should be source-indexed:

```php
$source_post_types = $this->cached_post_types[ $source ];
```

… but ideally its lifetime should be that of a request.

No?

includes/rest-api/class-acf-rest-types-endpoint.php
```diff
@@ -14,19 +14,174 @@
...
+				return new WP_Error(
+					'rest_post_type_invalid',
+					__( 'Invalid post type.', 'secure-custom-fields' ),
+					array( 'status' => 404 )
+				);
+			}
+		}
+		// For collection requests, we don't need to add any filter here
+		// as clean_types_response will handle removing null values from the response
+		// and filter_post_type will handle individual filtering
```
mcsf: In light of this, why even check for `$is_collection`? Seems like a distraction, no?

includes/rest-api/class-acf-rest-types-endpoint.php
```diff
@@ -14,19 +14,174 @@
...
+
+		if ( 'core' === $source || 'other' === $source ) {
+			$all_post_types = get_post_types( array( '_builtin' => true ), 'objects' );
+			foreach ( $all_post_types as $post_type ) {
+				$core_types[] = $post_type->name;
+			}
+		}
+
+		if ( 'scf' === $source || 'other' === $source ) {
+			$scf_types = array();
```
mcsf: Why the reassignment?

mcsf (commented):
I didn't comb through every little detail, but overall seems pretty good! I just left a concern about caching.

includes/rest-api/class-acf-rest-types-endpoint.php
priethor: You are right! Instead of source-indexing it, which was something I considered from the start and wanted to avoid, I've opted for a cleaner solution in 555217b: clear the cache at the end of each request, so that it is truly stateless.

priethor: @mcsf I completely removed the cache in a525023  to make it TRULY stateless.

includes/rest-api/class-acf-rest-types-endpoint.php
```diff
@@ -123,4 +253,116 @@ private function get_field_schema() {
...
+	 *
+	 * @since SCF 6.5.0
+	 *
+	 * @param array           $response The response data.
+	 * @param WP_REST_Server  $server   The REST server instance.
+	 * @param WP_REST_Request $request  The original request.
+	 * @return array            The filtered response data.
+	 */
+	public function clean_types_response( $response, $server, $request ) {
+		if ( strpos( $request->get_route(), '/wp/v2/types' ) !== 0 ) {
```
mcsf: This match is too eager, catching any word starting with `types`. Could be a regex pattern similar to one already used in this diff: `preg_match( '#^/wp/v2/types/([^/]+)$#'`. Drive-by questions:

- is there a REST utility for this?
- can we assume a trailing slash will always be there, or do we need to account for `/types$/`?

includes/rest-api/class-acf-rest-types-endpoint.php
```diff
@@ -27,6 +27,135 @@ class SCF_Rest_Types_Endpoint {
...
+	private function get_source_post_types( $source ) {
+
+		$core_types = array();
+		$scf_types  = array();
+
+		if ( 'core' === $source || 'other' === $source ) {
+			$all_post_types = get_post_types( array( '_builtin' => true ), 'objects' );
+			foreach ( $all_post_types as $post_type ) {
+				$core_types[] = $post_type->name;
+			}
```
mcsf, line 126: There's `wp_list_pluck`, though I don't know if it performs worse. :)

Ideally, there would be `array_column`, but AFAIK that operates on items of type array, not object.

(This is an FYI, no need to change unless you'd prefer to.)

includes/rest-api/class-acf-rest-types-endpoint.php
```diff
@@ -27,6 +27,135 @@ class SCF_Rest_Types_Endpoint {
...
+		}
+
+		if ( 'scf' === $source || 'other' === $source ) {
+			// Get SCF-managed post types
+			if ( function_exists( 'acf_get_internal_post_type_posts' ) ) {
+				$scf_managed_post_types = acf_get_internal_post_type_posts( 'acf-post-type' );
+				foreach ( $scf_managed_post_types as $scf_post_type ) {
+					if ( isset( $scf_post_type['post_type'] ) ) {
```
mcsf: Asking as SCF-ignorant: when would there be no `post_type` in an item in a collection returned by a function named _acf_get_internal_post_type__**posts**_?

includes/rest-api/class-acf-rest-types-endpoint.php
mcsf: I like this too; why not? :)

includes/rest-api/class-acf-rest-types-endpoint.php
```diff
@@ -123,4 +253,116 @@ private function get_field_schema() {
...
+	private function get_source_param_definition( $include_validation = false ) {
+		$param = array(
+			'description' => __( 'Filter post types by their source.', 'secure-custom-fields' ),
+			'type'        => 'string',
+			'enum'        => array( 'core', 'scf', 'other' ),
+			'required'    => false,
+		);
+
+		// Not needed for OpenAPI documentation
+		if ( $include_validation ) {
```
mcsf: Is it common practice to explicitly remove them? In other words, is the small complexity cost of this branching worth the benefit of not having these params in the route map? Also, if these params aren't needed and are undesirable, why doesn't the REST framework strip them for all?

mcsf (commented):
This looks much much better. Left some notes, still. Also, don't rely on my approval for merging! You all have the domain knowledge, not me. :)

includes/rest-api/class-acf-rest-types-endpoint.php
priethor: It's primarily a safety check for unvalidated raw data, since this function was only introduced in ACF 6.1, but the result from this function is actually validated (as opposed to `acf_get_raw_internal_post_type_posts()` so it should not be necessary even for extreme cases like corrupted/non-migrated data.

includes/rest-api/class-acf-rest-types-endpoint.php
priethor: Great questions. The context here is that this is a result of combining the two independent functions that shared the first part for the actual parameter registration and the OpenAPI docs, whereas the OpenApi docs only needed the first part.

To answer your questions, I haven't found examples of explicit removals, but I'm unsure about the efficiency: isn't a conditional check more efficient than a bunch of assignments or making the REST API strip them? In either case, the performance difference might be negligible, and the branching not worth it, as it decreases readability.

includes/rest-api/class-acf-rest-types-endpoint.php
priethor: > is there a REST utility for this?

Not that I found, but you are right, it's too open. I should update this one to use `preg_match` too, but working with both individual post types _and_ collections.

> can we assume a trailing slash will always be there, or do we need to account for /types$/?

No, we need to support with and without trailing slash

includes/rest-api/class-acf-rest-types-endpoint.php
mcsf: I don't really have a position here, mind you. :) My instinct is "why do more work?", but I'm aware that I'm almost bike-shedding here. Totally up to you whether to change anything here or not.

priethor:
Ok, merging at last. Thanks for all the thorough reviews 🚀

PR #132, opened 2025-05-14
Bug: Fix repeater inner fields not showing in Field Group admin page.

cbravobernal:
## What

A small typo was breaking the repeater field, it was not showing the fields.

Before:
Screenshot 2025-05-14 at 19 07 56

After:
Screenshot 2025-05-14 at 19 09 56

I will work on an E2E test as soon as possible. Bug was introduced in 93c1c8b4a77465922e57f062e09e10bb40f35f3f

### DIFF
--- a/includes/acf-hook-functions.php
+++ b/includes/acf-hook-functions.php
@@ -106,7 +106,7 @@ function _acf_apply_hook_variations() {
 		}
 
 		// Apply filters.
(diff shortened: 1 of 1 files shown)


priethor (approved):
Checked and debugged in real time with Carlos, works as expected 👍

PR #135, opened 2025-05-15
Fix: SCF Blocks are now forced into preview mode when editing a synced pattern.

cbravobernal:
## What

Before this PR. SCF blocks could be on "edit" mode inside synced patterns. Now they are forced to be in preview mode, which is the block that the user will see both in editor and in frontend.

ACF 6.4.1 backport.

## How to test
Before:

https://github.com/user-attachments/assets/13844576-5139-4191-80f0-b34a5c02af1c

After:

https://github.com/user-attachments/assets/a2ba2436-2eb5-411c-bb0e-62fcb7c6fcbe

### DIFF
--- a/assets/src/js/pro/_acf-blocks.js
+++ b/assets/src/js/pro/_acf-blocks.js
@@ -110,7 +110,7 @@ const md5 = require( 'md5' );
 	 * @return boolean
 	 */
 	function isSiteEditor() {
-		return typeof pagenow === 'string' && pagenow === 'site-editor';
+		return document.querySelectorAll( 'iframe[name="editor-canvas"]' ).length > 0;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


assets/src/js/pro/_acf-blocks.js
```diff
@@ -110,7 +110,7 @@ const md5 = require( 'md5' );
 	 * @return boolean
 	 */
 	function isSiteEditor() {
-		return typeof pagenow === 'string' && pagenow === 'site-editor';
+		return document.querySelectorAll( 'iframe[name="editor-canvas"]' ).length > 0;
```
priethor, line 113: Is this backported 1:1? Is there any other condition that makes this iframe check more accurate than just `wp.data.select('core/edit-site')` to know if we are in the Site Editor?

cbravobernal: Yep.

cbravobernal:
I'm going to merge the PR for early "prod" testing.

Issue #154, opened 2025-05-20
Support Custom Field Editing in Editor Sidebar


priethor:
This project introduces a new beta feature for Secure Custom Fields (SCF) that moves field group elements from their traditional location in the main content area to the WordPress editor sidebar. The goal is to create a more streamlined and modern editing experience by allowing users to view and edit custom fields directly within the Block Editor’s sidebar, similar to how core WordPress features like featured images and categories appear. This approach aligns with SCF’s broader initiative to modernize admin screens, reduce clutter in the main content area, and provide a more intuitive interface for managing structured content.

### Additional Context

- **Risks:**  
  - There may be UX challenges in presenting complex or numerous custom fields in the limited sidebar space, potentially leading to clutter or confusion.
  - Reimplementing each field type leveraging `@wordpress/components` is a huge undertaking; we should follow the 80-20 rule and consider alternatives for more niche field types.

- **Dependencies:**  
  - May require enhancements in WordPress core for advanced custom field types and improved sidebar extensibility.
  - May require improvements to `@wordpress/components` to handle some field types.

- **Other Considerations:**  
  - This feature will be introduced as an opt-in beta that administrators can enable.

PR #169, opened 2025-05-23
Bindings: Try to integrate SCF custom fields in the Attributes panel

gziolo:
Related to https://github.com/WordPress/secure-custom-fields/issues/154.

This is a proof of concept that explores including registered custom SCF fields in the Attributes panel, which is currently limited to show only fields from the Post Meta binding source.

Screenshot 2025-05-23 at 14 45 34

This change alone allows connecting in the editor fields registered with SCF with all core blocks that support Block Bindings.

## Approach

This prototype uses the [Proxy object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) to override the `getBlockBindingsSources` method in `wp.blocks` global. Proxy is necessary because properties inside `wp.blocks` are set as readonly, so that is an elegant way to tap into the existing system. 

I tried replicating the same Attributes panel by duplicating code from Gutenberg, but it uses some internal/private APIs that are difficult to mirror in a plugin. Therefore, I opted for an existing method that is used to construct the list of fields that can be connected with blocks. In the ideal world, some changes in Gutenberg would need to be applied to allow custom Block Bindings sources to define the list of fields to connect.

I haven’t explored that yet, but in the editor, users see the `SCF Fields` label in readonly mode without more details about the specific connected fields. It's probably caused by the fact that SCF Fields source is only registered on the server, so it doesn’t have the logic necessary to show something more specific like Post Meta does.

## Demo

https://github.com/user-attachments/assets/927f0e1e-eca0-4c07-8fe6-30b17bc1d618

### DIFF
--- a/includes/assets.php
+++ b/includes/assets.php
@@ -277,18 +277,14 @@ window.wp.blocks = new Proxy(originalBlocks, {
 		console.log("Intercepted getBlockBindingsSources", result);
 		if (result?.["acf/field"]) {
 				result["acf/field"]["getFieldsList"] = function() {
-					return {
-						isbn: {
-							label: "ISBN (SCF)",
-							value: "978-3-16-148410-0",
-							type: "string",
-						},
-						author: {
-							label: "Author (SCF)",
-							value: "John Doe",
-							type: "string",
-						},
-					};
+					return acf.getFields().reduce( (acc, { data, $el }) => {
+						acc[data.name] = {
+							label: $el.find("label").text(),
+							value: $el.find("input").val() || "",
+							type: data.type === "text" ? "string" : data.type,
+						};
+						return acc;
+					}, {} );
 				};
 		}
         return result;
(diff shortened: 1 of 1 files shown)


includes/assets.php
```diff
@@ -265,6 +265,37 @@ public function register_scripts() {
...
+  get(target, prop, receiver) {
+    if (prop === "getBlockBindingsSources") {
+      return function(...args) {
+        const result = target.getBlockBindingsSources(...args);
+		if (result?.["acf/field"]) {
+				result["acf/field"]["getFieldsList"] = function() {
+					return acf.getFields().reduce( (acc, { data, $el }) => {
+						acc[data.name] = {
+							label: $el.find("label").text(),
+							value: $el.find("input").val() || "",
```
cbravobernal, line 282: I guess this won't work for fields different that the text one. (For example the textarea).

gziolo: Sure, it would have to cover all types of input from different field types. I covered the happy path for the text field as it's the only one supported in the Attributes panel anyway.

gziolo:
The work continues to open the interface through the Attributes panel in Gutenberg and later WordPress 6.9. In addition to that, SCF now has a custom built-in integration with bindings.

PR #175, opened 2025-06-06
Block Bindings: Add editor preview to acf-field source.

cbravobernal:
## What

Right now, the experience with block bindings in SCF is not the best. The image block binding is broken. There is an extra warning from array to string conversion, and all the block binding place holders are just the default placeholder. (With the SCF Fields name)

https://github.com/user-attachments/assets/b5cdaadf-831d-4591-9ec0-2ece8bcab864

This PR enhances the experience on the editor by using the new Editor APIs. Now only we use the getValues one. Preventing from edition, but https://github.com/WordPress/secure-custom-fields/pull/173 already allows to edit them too.

I need some more testing with different field types for editing. But with this PR, we cover all block bindings supported blocks and their attributes with field types like select, checkbox, url, text field, textareas and numbers.

## Why

To enhance the editor experience with block bindings and custom fields.

## Screenshare and how to test.

https://github.com/user-attachments/assets/555c592a-fcc4-4749-aee2-3afb9fe9aa0a

Create a custom post type, a field group, some fields.

Use the code to link those fields with their respective blocks, sharing an example so you can copy-paste:

**Heading**
```
<!-- wp:heading {"level":4,"metadata":{"bindings":{"content":{"source":"acf/field","args":{"key":"videogame_title"}}}}} -->
<h4 class="wp-block-heading"></h4>
<!-- /wp:heading -->
```

```
<!-- wp:heading {"level":4,"metadata":{"bindings":{"content":{"source":"acf/field","args":{"key":"hours_played"}}}}} -->
<h4 class="wp-block-heading"></h4>
<!-- /wp:heading -->
```

**Image**

```
<!-- wp:image {"metadata":{"bindings":{"id":{"source":"acf/field","args":{"key":"videogame_image"}},"url":{"source":"acf/field","args":{"key":"videogame_image"}},"alt":{"source":"acf/field","args":{"key":"videogame_image"}},"title":{"source":"acf/field","args":{"key":"videogame_image"}}}}} -->
<figure class="wp-block-image"><img src="#" alt=""/></figure>
<!-- /wp:image -->
```

**Button**
```
<!-- wp:buttons -->
<div class="wp-block-buttons"><!-- wp:button {"metadata":{"bindings":{"text":{"source":"acf/field","args":{"key":"text"}},"url":{"source":"acf/field","args":{"key":"igdb_url"}},"linkTarget":{"source":"acf/field","args":{"key":"igdb_target"}},"rel":{"source":"acf/field","args":{"key":"igdb_rel"}}}}} -->
<div class="wp-block-button"><a class="wp-block-button__link wp-element-button">Binded Button</a></div>
<!-- /wp:button --></div>
<!-- /wp:buttons -->
```

### DIFF
--- a/assets/src/js/bindings/sources.js
+++ b/assets/src/js/bindings/sources.js
@@ -51,10 +51,15 @@ registerBlockBindingsSource( {
 
 				const fieldValue = getFieldValue( fields, fieldName );
 				if ( typeof fieldValue === 'object' && fieldValue !== null ) {
-					result[ attribute ] =
-						( fieldValue[ attribute ] ??
-							( attribute === 'content' && fieldValue.url ) ) ||
-						'';
+					let value = '';
+
+					if ( fieldValue[ attribute ] ) {
+						value = fieldValue[ attribute ];
+					} else if ( attribute === 'content' && fieldValue.url ) {
+						value = fieldValue.url;
+					}
+
+					result[ attribute ] = value;
 				} else if ( typeof fieldValue === 'number' ) {
 					if ( attribute === 'content' ) {
 						result[ attribute ] = fieldValue.toString() || '';
--- a/assets/src/js/bindings/sources.js
+++ b/assets/src/js/bindings/sources.js
@@ -1,0 +1,79 @@
+/**
+ * WordPress dependencies.
+ */
+import { __ } from '@wordpress/i18n';
+import { registerBlockBindingsSource } from '@wordpress/blocks';
+import { store as coreDataStore } from '@wordpress/core-data';
+
+/**
+ * Get the value of a specific field from the ACF fields.
+ *
+ * @param {Object} fields The ACF fields object.
+ * @param {string} fieldName The name of the field to retrieve.
+ * @returns {string} The value of the specified field, or undefined if not found.
+ */
+const getFieldValue = ( fields, fieldName ) => fields?.acf?.[ fieldName ];
+
+const resolveImageAttribute = ( imageObj, attribute ) => {
+	if ( ! imageObj ) return '';
+	switch ( attribute ) {
+		case 'url':
+		case 'content':
+			return imageObj.source_url;
+		case 'alt':
+			return imageObj.alt_text || '';
+		case 'title':
+			return imageObj.title?.rendered || '';
+		default:
+			return '';
+	}
+};
+
+registerBlockBindingsSource( {
+	name: 'acf/field',
+	label: 'SCF Fields',
+	getValues( { context, bindings, select } ) {
+		const { getEditedEntityRecord, getMedia } = select( coreDataStore );
+		let fields =
+			context?.postType && context?.postId
+				? getEditedEntityRecord(
+						'postType',
+						context.postType,
+						context.postId
+				  )
+				: undefined;
+		const result = {};
+		debugger;
+
+		Object.entries( bindings ).forEach(
+			( [ attribute, { args } = {} ] ) => {
+				const fieldName = args?.key;
+
+				const fieldValue = getFieldValue( fields, fieldName );
+				if ( typeof fieldValue === 'object' && fieldValue !== null ) {
+					result[ attribute ] =
+						( fieldValue[ attribute ] ??
+							( attribute === 'content' && fieldValue.url ) ) ||
+						'';
+				} else if ( typeof fieldValue === 'number' ) {
+					if ( attribute === 'content' ) {
+						result[ attribute ] = fieldValue.toString() || '';
+					} else {
+						const imageObj = getMedia( fieldValue );
+						result[ attribute ] = resolveImageAttribute(
+							imageObj,
+							attribute
+						);
+					}
+				} else {
+					result[ attribute ] = fieldValue || '';
+				}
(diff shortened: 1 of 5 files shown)


assets/src/js/bindings/sources.js
```diff
@@ -0,0 +1,79 @@
...
+		Object.entries( bindings ).forEach(
+			( [ attribute, { args } = {} ] ) => {
+				const fieldName = args?.key;
+
+				const fieldValue = getFieldValue( fields, fieldName );
+				if ( typeof fieldValue === 'object' && fieldValue !== null ) {
+					result[ attribute ] =
+						( fieldValue[ attribute ] ??
+							( attribute === 'content' && fieldValue.url ) ) ||
+						'';
```
ockham: The expression is a bit hard to read (at least for me, since I'm not totally sure about the order of precedence of the `??` and `||` here). Maybe turn at least one of them into an `if` clause?

includes/Blocks/Bindings.php
```diff
@@ -78,15 +79,42 @@ public function get_value( array $source_attrs, \WP_Block $block_instance, strin
...
-			$value = $field['value'];
+			switch ( $attribute_name ) {
+				case 'id':
+					// The value is the field value.
+					$value = $field['value']['id'] ?? '';
+					break;
+				case 'alt':
+					// The label is the field label.
+					$value = $field['value']['alt'] ?? '';
+					break;
```
ockham: Could do

```suggestion
				case 'id':
				case 'alt':
				case 'title':
					// The value is in the field of the same name.
					$value = $field['value'][ $attribute_name ] ?? '';
					break;
```

includes/Blocks/Bindings.php
```diff
@@ -78,15 +79,42 @@ public function get_value( array $source_attrs, \WP_Block $block_instance, strin
 				}
 			}
 
-			$value = $field['value'];
+			switch ( $attribute_name ) {
```
ockham, line 82: The code in this `switch` statement makes some assumptions about various bound block attributes. But those could be different for other block types, no? Do we need to add an extra check for the block type to make sure we're processing attributes correctly? (E.g. the code for the `alt`, `url`, or `title` attributes here seems specific to the `core/image` block. But what if a third-party block has an attribute of the same name? It might require different processing, no?)

assets/src/js/bindings/sources.js
cbravobernal: Done in https://github.com/WordPress/secure-custom-fields/pull/175/commits/e7b46a5674d21438d5ed7f225f9a4967fef393fc

includes/Blocks/Bindings.php
cbravobernal: Done in https://github.com/WordPress/secure-custom-fields/pull/175/commits/e7b46a5674d21438d5ed7f225f9a4967fef393fc

includes/Blocks/Bindings.php
cbravobernal: Right now, the only supported blocks and attributes are these:

Supported Blocks | Supported Attributes
-- | --
Image | url, alt, title
Paragraph | content
Heading | content
Button | url, text, linkTarget, rel

For that reason, we can afford to assume that a blocks having those attributes are an image one. We still haven't decided how to support future blocks and their attributes, so this code is expected to change in a "future".

includes/Blocks/Bindings.php
```diff
@@ -78,15 +79,36 @@ public function get_value( array $source_attrs, \WP_Block $block_instance, strin
...
-			}
+					if ( is_array( $value ) ) {
+						$value = wp_json_encode( $value );
+					}
 
-			// If we're not a scalar we'd throw an error, so return early for safety.
-			if ( ! is_scalar( $value ) ) {
-				$value = null;
+					// Ensure we're returning a scalar value.
+					if ( ! is_scalar( $value ) && null !== $value ) {
```
priethor: This is mutually exclusive with the `is_array` check earlier; it could be an `else if` statement.

assets/src/js/bindings/sources.js
```diff
@@ -0,0 +1,84 @@
...
+		let fields =
+			context?.postType && context?.postId
+				? getEditedEntityRecord(
+						'postType',
+						context.postType,
+						context.postId
+				  )
+				: undefined;
+		const result = {};
+		debugger;
```
priethor: Leftover debugging code.
```suggestion
```

cbravobernal: nice catch 😆

cbravobernal: Fixed in https://github.com/WordPress/secure-custom-fields/pull/175/commits/61f7db5b6b2a7442c67bd9dd94ea5ac0028e7fe5

includes/Blocks/Bindings.php
cbravobernal: Fixed in https://github.com/WordPress/secure-custom-fields/pull/175/commits/61f7db5b6b2a7442c67bd9dd94ea5ac0028e7fe5

gziolo:
Nice one!

PR #176, opened 2025-06-12
Connect block attributes with custom fields via UI

cbravobernal:
## What

⚠️ It needs to be loaded as a beta feature in their respective admin panel.
⚠️ Pattern overrides cannot be set up connected to fields. They are still not compatible on Core.
⚠️ There is some duplication both in code and UI with Core.

The PR allows to connect block attributes with custom fields via UI. Is still experimental, as I have not yet tested with any field type that can be created. In a future PR we can add an advanced view, using DataViews.

The image block will only search for image type fields, and all their attributes can be bound at once, just selecting the image field.

The paragraph, heading, and button blocks will only work with an initial set of items set in this variable:

```js
const BLOCK_BINDINGS_RELATED_FIELD_TYPES = {
	'core/paragraph': {
		content: [ 'text', 'textarea', 'date_picker', 'number', 'range' ],
	},
	'core/heading': {
		content: [ 'text', 'textarea', 'date_picker', 'number', 'range' ],
	},
	'core/image': {
		id: [ 'image' ],
		url: [ 'image' ],
		title: [ 'image' ],
		alt: [ 'image' ],
	},
	'core/button': {
		url: [ 'url' ],
		text: [ 'text', 'checkbox', 'select', 'date_picker' ],
		linkTarget: [ 'text', 'checkbox', 'select' ],
		rel: [ 'text', 'checkbox', 'select' ],
	},
};
```

## Screenshare

https://github.com/user-attachments/assets/d8c55adb-04e4-413a-aaba-378b17661832


ockham:
I think there's a minor UX glitch with regard to the tools panel and link/unlink button:

connect-fields

AFAIU, we should change two things here:

- The panel's heading shouldn't change based on the link/unlink button; it shouldn't describe the link/unlink option, but the nature of the panel's controls (so probably "Attributes").
- The panel's "vertical three dots" menu shouldn't change contextually; it should only apply to the whole Attributes panel, not its individual rows.

(This is based on how I observed the UX pattern to work e.g. for the existing margin panel in the style tab of the block inspector.)

cbravobernal:
> * The panel's "vertical three dots" menu shouldn't change contextually; it should only apply to the whole Attributes panel, not its individual rows.

But, in that case, if you have the four of them unlinked, and want to reset only the image title or alt, you will need to reset every of them at once and set one by one again, right?

It's better to keep consistency, but as we change the number of toolsitem, that is the default behaviour. I'll take a look at how is implemented for margin and styles.

Thanks @ockham!

assets/src/js/bindings/sources.js
```diff
@@ -59,7 +69,7 @@ registerBlockBindingsSource( {
 					}
 
 					result[ attribute ] = value;
-				} else if ( typeof fieldValue === 'number' ) {
+				} else if ( 'number' === typeof fieldValue ) {
```
ockham: Would it make sense to replace this with

```suggestion
				} else if ( 'number' === fieldType ) {
```

now that we have `fieldType` available?

cbravobernal: I refactored the entire part 😅 

Edit: Done! I think it was a different file.

cbravobernal:
Let's see copilot reviewing copilot code generated.

assets/src/js/bindings/sources.js
```diff
@@ -69,6 +80,12 @@ registerBlockBindingsSource( {
 							attribute
 						);
 					}
+				} else if ( 'date_picker' === fieldType && fieldValue ) {
```
cbravobernal: Let's check that fieldValue existence.

cbravobernal:
> I think there's a minor UX glitch with regard to the tools panel and link/unlink button:
> 
> connect-fields [ connect-fields ](https://private-user-images.githubusercontent.com/96308/462444107-2a64963f-80c3-422f-8669-aeeea09b905a.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NTIwNTE1NjcsIm5iZiI6MTc1MjA1MTI2NywicGF0aCI6Ii85NjMwOC80NjI0NDQxMDctMmE2NDk2M2YtODBjMy00MjJmLTg2NjktYWVlZWEwOWI5MDVhLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTA3MDklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUwNzA5VDA4NTQyN1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWJiMDAzM2ZjNWQyNGMyNWEzZGNhMzU0YTVlNGMyN2QzYzM4MmMyZWI0OWQzNDZkYjQ1YWJiODk1ZWJlYTVjZTcmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.-2i-GNek3t7_RXeFLqugbsVqx6RwZDQBVTy1K0CmMYI) [ ](https://private-user-images.githubusercontent.com/96308/462444107-2a64963f-80c3-422f-8669-aeeea09b905a.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NTIwNTE1NjcsIm5iZiI6MTc1MjA1MTI2NywicGF0aCI6Ii85NjMwOC80NjI0NDQxMDctMmE2NDk2M2YtODBjMy00MjJmLTg2NjktYWVlZWEwOWI5MDVhLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTA3MDklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUwNzA5VDA4NTQyN1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWJiMDAzM2ZjNWQyNGMyNWEzZGNhMzU0YTVlNGMyN2QzYzM4MmMyZWI0OWQzNDZkYjQ1YWJiODk1ZWJlYTVjZTcmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.-2i-GNek3t7_RXeFLqugbsVqx6RwZDQBVTy1K0CmMYI)
> 
> AFAIU, we should change two things here:
> 
> * The panel's heading shouldn't change based on the link/unlink button; it shouldn't describe the link/unlink option, but the nature of the panel's controls (so probably "Attributes").
> * The panel's "vertical three dots" menu shouldn't change contextually; it should only apply to the whole Attributes panel, not its individual rows.
> 
> (This is based on how I observed the UX pattern to work e.g. for the existing margin panel in the style tab of the block inspector.)

I finally removed the option of using that button link. Images will connect only to image-type fields.

ockham:
If I create a new CPT and corresponding field group, but forget to enable the "Allow Access to Value in Editor UI" toggle in a field's "Presentation" tab, I'm still able to connect blocks to that field in the editor. However, it doesn't show up on the frontend:

connect-fields-toggle

It works on the frontend once I enable those toggles.

I guess the solution is to disallow connecting to a field in the editor if its "Allow Access to Value in Editor UI" toggle is disabled?

gziolo:
Removing selected connection doesn't work properly:

Screenshot 2025-07-09 at 13 44 40


Screenshot 2025-07-09 at 13 43 49

The editor remains in this weird state where it looks like the field is still connected with SCF field.

gziolo:
For some reason, the panel moves in the sidebar after selecting the value:

Screenshot 2025-07-09 at 13 46 27

Screenshot 2025-07-09 at 13 45 56

gziolo:
The experience for the Button block is less fluid. Should we limit it to mapping to the text attribute and leave the advanced connection to the Attributes panel?

Screenshot 2025-07-09 at 13 49 24

Maybe we could limit it to URL and text?

The filtering for the URL looks pretty nice:

Screenshot 2025-07-09 at 13 53 04

cbravobernal:
> For some reason, the panel moves in the sidebar after selecting the value:
> 
> Screenshot 2025-07-09 at 13 46 27 Screenshot 2025-07-09 at 13 45 56

That happened too in Gutenberg with the tab. It maybe a hook-related thing. I'll take a look at it.

assets/src/js/bindings/sources.js
```diff
@@ -24,54 +49,133 @@ const resolveImageAttribute = ( imageObj, attribute ) => {
...
+ * @param {Object} fieldConfig The field configuration object.
+ * @returns {string} The formatted date string.
+ */
+const handleDateFieldValue = ( fieldValue, fieldConfig ) => {
+	if ( ! fieldValue ) {
+		return '';
+	}
+
+	return dateI18n( fieldConfig?.display_format, fieldValue ) || '';
+};
```
ockham: I'd be interested in making `date_picker` fields work with Core's Date block (that I'm currently [working on](https://github.com/WordPress/gutenberg/pull/70585)). That block will allow binding to its `datetime` attribute, which will be using [this format](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time#valid_datetime_values). The user can use the block inspector to set the format that the date is rendered with (this is a feature of the Date block that already exists).

However, AFAICS, the SCF source will always return the date using the `display_format` that's set in SCF options, meaning that the Date block might receive the date in a format that it won't understand ❌ 

In this particular case, one solution would be to allow binding the Date block's `format` attribute to a source, and to have SCF expose its format. This could be a fairly elegant solution, as it would allow both using SCF's format, but also setting a custom format from within the Date block. (We would need to allow unlinking individual bound attributes, however.)

This might work well enough in the case of date fields, but might not scale to other cases. In the bigger picture, this is a question of who decides the formatting of a given piece of information. An alternative approach would be for SCF to always just provide a standardized "raw" format, and to leave formatting up to the individual block that it is bound to. IMO, this makes some sense, as it encapsulates the data layer in SCF, and leaves the formatting to the presentational layer (i.e. the blocks). However, it would mean that SCF's own format settings will be ignored, and some data won't look pretty when bound to a more "primitive" block (e.g. binding `datetime` to a paragraph block).

I was wondering how the question of formatting was previously considered in the context of Block Bindings. For example, so far, all examples of Block Bindings use `args: { key: ... }` to select a given source field for binding; was the idea that `format` could be supplied as an additional arg? (E.g. `args: { key: date, format: 'M j, Y' }`?

@cbravobernal @gziolo 

---

We probably don't have to answer this for this PR, but it might be good to keep in mind for our next steps.

assets/src/js/bindings/sources.js
```diff
@@ -24,54 +49,133 @@ const resolveImageAttribute = ( imageObj, attribute ) => {
...
+ */
+const handleNumericFieldValue = ( fieldValue, attribute, getMedia ) => {
+	if ( attribute === 'content' ) {
+		return fieldValue.toString() || '';
+	}
+
+	// For image fields or numeric values, try to resolve as media
+	const imageObj = getMedia( fieldValue );
+	return resolveImageAttribute( imageObj, attribute );
+};
```
ockham: I'd love if we could restructure this a bit. IMO, it feels a bit backwards that when given a numeric value, we have to guess if it's an image ID, when we have higher-level information about field types available at the callsite. (A similar note applies to `handleObjectFieldValue`.)

I'll quickly try out locally what that would look like.

assets/src/js/bindings/sources.js
cbravobernal: >  However, it would mean that SCF's own format settings will be ignored, and some data won't look pretty when bound to a more "primitive" block (e.g. binding datetime to a paragraph block).

I guess we could auto-convert paragraphs bounded to a datetime field to that date block to avoid that. But then it would not be compatible with future "shortblocks". My opinion is that the format value defined on the field has been widely used, so it should be the default one, and then the block should be capable to override that format option.

That way we wouldn't break backwards compatibility neither. We could for sure update the source with more args. The key one was a convention from the beginning of the project.

cbravobernal:
> The experience for the Button block is less fluid. Should we limit it to mapping to the text attribute and leave the advanced connection to the Attributes panel?

Sure, my opinion is that the `rel` and `target` options should be in the block toolbar or the block sidebar, rather than being bindings (for UI based blocks). I would keep those bindings just for more complex cases (attributes based on external sources or php conditionals).

assets/src/js/bindings/sources.js
cbravobernal: Please! Totally agree.

assets/src/js/bindings/sources.js
```diff
@@ -24,54 +49,133 @@ const resolveImageAttribute = ( imageObj, attribute ) => {
...
+	// Check if the field value has the exact attribute property
+	if ( fieldValue.hasOwnProperty( attribute ) && fieldValue[ attribute ] ) {
+		return fieldValue[ attribute ];
+	}
+
+	// Special fallback: if we're looking for 'content' and no content property exists (or is falsy),
+	// but there's a 'url' property, use that instead
+	if ( attribute === 'content' && fieldValue.url ) {
+		return fieldValue.url;
+	}
```
ockham: Is this for images?

assets/src/js/bindings/sources.js
ockham: This is more or less what I have in mind: At the top level, a `switch` for all possible SCF field types. It's still incomplete since I'm not yet totally familiar with all the possible ways they are mapped to block attributes.

```js
	switch ( fieldType ) {
		case 'text':
		case 'textarea':
			return fieldValue || '';
		case 'number':
			if ( attribute === 'content' ) {
				return fieldValue.toString() || '';
			}
			break;
		case 'image':
			// fieldValue is a (numeric) image ID.
			const imageObj = getMedia( fieldValue );
			return resolveImageAttribute( imageObj, attribute );
		case 'date_picker':
			if ( ! fieldValue ) {
				return '';
			}
			return dateI18n( fieldConfig?.display_format, fieldValue ) || '';
	}
```

cbravobernal:
> If I create a new CPT and corresponding field group, but forget to enable the "Allow Access to Value in Editor UI" toggle in a field's "Presentation" tab, I'm still able to connect blocks to that field in the editor. However, it doesn't show up on the frontend:
> 
> ......
> It works on the frontend once I enable those toggles.
> 
> I guess the solution is to disallow connecting to a field in the editor if its "Allow Access to Value in Editor UI" toggle is disabled?

Fixed in https://github.com/WordPress/secure-custom-fields/pull/176/commits/89d69ffbef77c6201a9b370143ad49c56f52ae97
 It will only show fields with the toggle enabled.

assets/src/js/bindings/sources.js
ockham: Done in https://github.com/WordPress/secure-custom-fields/pull/176/commits/2c96d621ab748e9849bf8410a9db3a57f0192663.

ockham:
BTW the `date_picker` field currently shows a date in the editor (when bound to a paragraph) that's different from what's set in SCF 🙈 

image

My hunch is that it has to do with timezones -- one of them probably uses UTC, and the other one the local timezone 😬

assets/src/js/bindings/sources.js
cbravobernal: Image doesn't have 'content' as a bindable attribute, lets change that.
Updated in https://github.com/WordPress/secure-custom-fields/pull/176/commits/60976f670c8466ded21a39b091f95243afbdac86

cbravobernal:
> BTW the `date_picker` field currently shows a date in the editor (when bound to a paragraph) that's different from what's set in SCF 🙈
> 
> image
> My hunch is that it has to do with timezones -- one of them probably uses UTC, and the other one the local timezone 😬

Ouch. 

> BTW the `date_picker` field currently shows a date in the editor (when bound to a paragraph) that's different from what's set in SCF 🙈
> 
> image
> My hunch is that it has to do with timezones -- one of them probably uses UTC, and the other one the local timezone 😬

Should we just take the formatted value from the server and print it as a string?

ockham:
> Should we just take the formatted value from the server and print it as a string?

Maybe 🤔 

Plus I guess I need to file a bug report against Gutenberg, since it seems that `dateI18n` (JS) behaves differently from [`date_i18n`](https://developer.wordpress.org/reference/functions/date_i18n/) (PHP).

cbravobernal:
> > Should we just take the formatted value from the server and print it as a string?
> 
> Maybe 🤔
> 
> Plus I guess I need to file a bug report against Gutenberg, since it seems that `dateI18n` (JS) behaves differently from [`date_i18n`](https://developer.wordpress.org/reference/functions/date_i18n/) (PHP).

I have a prototype working, but is adding all values parsed by SCF to the types endpoint, ending with maybe too much information.

I will instead try to extend acf object data to include the formatted value.

ockham:
Per discussion via DM, I've pushed a fix for the date issue: https://github.com/WordPress/secure-custom-fields/pull/176/commits/718af07d93ded3c12b3ec733b064bb767c77687b

assets/src/js/bindings/block-editor.js
```diff
@@ -0,0 +1,407 @@
...
+		const fields = useMemo(
+			() =>
+				fieldsGroups?.reduce( ( acc, fieldGroup ) => {
+					const groupFields =
+						fieldGroup.fields?.map( ( field ) => ( {
+							...field,
+							fieldGroupTitle: fieldGroup.title,
+							name: field.name,
+							label: field.label,
+							value: field.value,
```
ockham: I've inspected this, and `value` seems to be always `undefined` for me 🤔

assets/src/js/bindings/block-editor.js
```diff
@@ -0,0 +1,407 @@
...
+
+		const currentBindings = props.attributes?.metadata?.bindings || {};
+
+		const fields = useMemo(
+			() =>
+				fieldsGroups?.reduce( ( acc, fieldGroup ) => {
+					const groupFields =
+						fieldGroup.fields?.map( ( field ) => ( {
+							...field,
+							fieldGroupTitle: fieldGroup.title,
```
ockham: Do we need `fieldGroupTitle`? We don't seem to be using it anywhere...

assets/src/js/bindings/block-editor.js
```diff
@@ -0,0 +1,407 @@
...
+							name: field.name,
+							label: field.label,
+							value: field.value,
+						} ) ) || [];
+
+					return [ ...acc, ...groupFields ];
+				}, [] ) || [],
+			[ fieldsGroups ]
+		);
```
ockham: If we don't need `fieldGroupTitle` (see https://github.com/WordPress/secure-custom-fields/pull/176/files#r2197468409), then I think we can simplify this as

```suggestion
		const fields = useMemo(
			() =>
				fieldsGroups.flatMap( ( fieldGroup ) => fieldGroup.fields ) ||
				[],
			[ fieldsGroups ]
		);
```

assets/src/js/bindings/block-editor.js
```diff
@@ -0,0 +1,407 @@
...
+				// Filter fields to only include those that match the allowed types for this block
+				return fields
+					.filter( ( field ) =>
+						uniqueFieldTypes.includes( field.type )
+					)
+					.map( ( field ) => ( {
+						value: field.name,
+						label: field.label,
+						fieldType: field.type,
+					} ) );
```
ockham: I don't quite get why we're `map`ping here. Seems like we're only selecting three properties (`value`, `label`, `type`), and renaming `type` to `fieldType` (_edit:_ and `name` to `value`). But do we even need to do that? (Same below on the `else` branch.)

assets/src/js/bindings/block-editor.js
cbravobernal: Sorry, the code got messed up, refactored in https://github.com/WordPress/secure-custom-fields/pull/176/commits/866bb43d399579ab4b23fa1549b3191714f606bc

assets/src/js/bindings/block-editor.js
cbravobernal: Sorry, the code got messed up, refactored in https://github.com/WordPress/secure-custom-fields/pull/176/commits/866bb43d399579ab4b23fa1549b3191714f606bc

assets/src/js/bindings/block-editor.js
cbravobernal: Sorry, the code got messed up, refactored in https://github.com/WordPress/secure-custom-fields/commit/866bb43d399579ab4b23fa1549b3191714f606bc

cbravobernal:
> Removing selected connection doesn't work properly:
> 
> .....
> The editor remains in this weird state where it looks like the field is still connected with SCF field.

This one has been fixed.

assets/src/js/bindings/block-editor.js
cbravobernal: Same, fixed in https://github.com/WordPress/secure-custom-fields/pull/176/commits/fc5ceff69e441654ddc2a9d0f768d457f2e4aa0c

cbravobernal:
> > For some reason, the panel moves in the sidebar after selecting the value:
> > Screenshot 2025-07-09 at 13 46 27 Screenshot 2025-07-09 at 13 45 56
> 
> That happened too in Gutenberg with the tab. It maybe a hook-related thing. I'll take a look at it.

Fixed in https://github.com/WordPress/secure-custom-fields/pull/176/commits/4a0aeaffb4d3fcd4fe4921bfcc23d77a433576ff

cbravobernal:
Let's merge it and iterate it if needed.

PR #179, opened 2025-06-17
Enhancement: Create new options when editing a fields value on Selector.

cbravobernal:
## What

Allow creating new options when editing the field’s type select value.

Requirements:
 - Stylized UI (using Select2).
 - Multiple values field settings enabled.

The PR just adds the `tags` option to Select2 initialization if the the settings are enabled.

## Screencast

https://github.com/user-attachments/assets/dd6e96b5-8bfd-462c-84a2-1ae98c0e9171


cbravobernal:
I'm merging this one, as is a backport that we should not change the code much more.

PR #180, opened 2025-06-18
Set the Escaped HTML warning notice disabled by default

cbravobernal:
## What

This notice was introduced in ACF 6.2.5. Since 6.4.1 has been disabled by default. To maintain upstream compatibility we are doing the same here.

## Extra

Added a couple of aria-labels.


includes/admin/admin-internal-post-type.php
```diff
@@ -45,7 +45,6 @@ public function __construct() {
 			add_action( 'current_screen', array( $this, 'current_screen' ) );
 			add_action( 'save_post_' . $this->post_type, array( $this, 'save_post' ), 10, 2 );
 			add_action( 'wp_ajax_acf/link_field_groups', array( $this, 'ajax_link_field_groups' ) );
-			add_filter( 'post_updated_messages', array( $this, 'post_updated_messages' ) );
```
cbravobernal, line 48: Fixes the bb infinite loop.

PR #181, opened 2025-06-18
Fix – Blocks registered via acf_register_block_type() with a `parent` value of `null` no longer fail to register

ockham:
## What

If you register a block like this one.

```php
add_action('acf/init', 'my_acf_blocks_init');
function my_acf_blocks_init() {

    // Check function exists.
    if( function_exists('acf_register_block_type') ) {

        // Register a testimonial block.
        acf_register_block_type(array(
            'name'              => 'testimonial',
            'title'             => __('Testimonial 2'),
            'description'       => __('A custom testimonial block. 2'),
            'render_callback'   => function(){ echo 'hello from plugin';},
            'category'          => 'formatting',
	    'parent' => null
        ));
    }
}
```

Without the PR you get:
Screenshot 2025-06-19 at 14 46 31

Applying the patch you get the block as expected.
Screenshot 2025-06-19 at 14 47 09

### DIFF
--- a/includes/blocks.php
+++ b/includes/blocks.php
@@ -414,6 +414,16 @@ function acf_validate_block_type( $block ) {
 		$block['supports']['jsx'] = $block['supports']['__experimental_jsx'];
 	}
 
+	// Normalize block 'parent' setting.
+	if ( array_key_exists( 'parent', $block ) ) {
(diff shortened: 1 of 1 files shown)


cbravobernal:
Works as expected.

PR #183, opened 2025-06-19
Enhancement: Add Fields to Woo Subscriptions when using HPOS

cbravobernal:
## What

Now the user can add Fields to Woo Subscriptions when using HPOS.

## How

By updating the new WC_Order class, checking for the subscriptions page and adding the correspondant metaboxes.

## Screenshots

Field Group rule:
Screenshot 2025-06-19 at 12 12 45

Before the PR:
Screenshot 2025-06-19 at 12 10 31

After the PR:
Screenshot 2025-06-19 at 12 11 19

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -11,9 +11,7 @@ parameters:
 		- docs
 		- lang
 		- wordpress (?)
+		- includes/forms/WC_Order.php
 	bootstrapFiles:
 		- vendor/php-stubs/woocommerce-stubs/woocommerce-stubs.php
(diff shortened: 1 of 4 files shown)


phpstan.neon
```diff
@@ -11,6 +11,7 @@ parameters:
 		- docs
 		- lang
 		- wordpress (?)
+		- includes/forms/WC_Order.php
```
cbravobernal, line 14: There are a couple of phpstubs that need to be updated before we can include this file.
https://github.com/stellarwp/php-stubs-woocommerce-subscriptions/issues/1

PR #191, opened 2025-07-01
Command Palette: Convert commands to sentence case

### LINKED ISSUE #190: Commands should be sentence case, not Title Case
As the title says, SCF-related commands are currently in Title Case. Even though there is a bit of a mix-match in core commands and many of them look Title Case, that is the result of commands including raw block names, which themselves are Title Case, but the recommended style for commands is sentence case.

See a mix of core and SCF commands:

Image

priethor:
Closes #190 

Converts the Command Palette commands to sentence case, aligning with core commands.

|Before|After|
|-|-|
|image|image|

### DIFF
--- a/includes/admin/admin-commands.php
+++ b/includes/admin/admin-commands.php
@@ -56,7 +56,7 @@ function acf_commands_init() {
 			$custom_post_types[] = array(
 				'name'         => $post_type['post_type'],
 				'all_items'    => $labels->all_items,
-				'add_new_item' => $labels->add_new_item,
+				'add_new_item' => ucfirst( strtolower( $labels->add_new_item ) ),
 				'icon'         => $post_type['menu_icon'] ?? '',
 				'label'        => $labels->name,
 				'id'           => $post_type['ID'],
--- a/assets/src/js/commands/admin-commands.js
+++ b/assets/src/js/commands/admin-commands.js
@@ -32,7 +32,7 @@ const registerAdminCommands = () => {
 	const commands = [
 		{
 			name: 'field-groups',
-			label: __( 'Field Groups', 'secure-custom-fields' ),
+			label: __( 'Field groups', 'secure-custom-fields' ),
 			url: 'edit.php',
 			urlArgs: { post_type: 'acf-field-group' },
 			icon: layout,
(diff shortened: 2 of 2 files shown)


includes/admin/admin-commands.php
```diff
@@ -56,7 +56,7 @@ function acf_commands_init() {
 			$custom_post_types[] = array(
 				'name'         => $post_type['post_type'],
 				'all_items'    => $labels->all_items,
-				'add_new_item' => $labels->add_new_item,
+				'add_new_item' => ucfirst( strtolower( $labels->add_new_item ) ),
```
priethor, line 59: As far as I understand, RTL languages don't have upper and lower case, so this shouldn't affect them, nor would they require different treatment; is that correct, @mcsf?

ockham: We shouldn't do this 🙈 Operations like this tend to break when applied across different languages, as they make implicit assumptions that don't always apply to all languages.

In this particular change, it breaks for German, where nouns are _always_ capitalized: "book" is "Buch", so "Neues buch hinzufügen" would be very wrong 😬 

Basically, localized strings should be left alone (which is also the reason why `$labels` has so many seemingly redundant strings) -- our best course of action is to continue to use `$labels->add_new_item`, and to handle the case of the strings in the individual translations.

priethor: Thanks Bernie, I suspected this could not play well in other languages 😅 

In this case, the `add_new_label` in English capitalizes both words, as seen in the `Add new X` button, whereas [core commands for adding new pages](https://github.com/WordPress/gutenberg/blob/1d0c43d99a972643b6d97efa3a33762e31759eef/packages/core-commands/src/admin-navigation-commands.js#L77C6-L77C34) don't reuse strings and have their own post/page-specific strings.

My best guess here would be to add a completely new label to avoid breaking others, which aligns with the seemingly redundant strings you mentioned. Do you have any other recommendations?

priethor:
Closing this PR as it is the wrong approach: these labels can't and shouldn't be tinkered with at the code level, as they are defined through the post type creation UI, which defaults them to `Add New X` but are ultimately up to the user to set up. If we want these sentences not to be Title case, we should change the default values, but we shouldn't override what the user configures.
image

includes/admin/admin-commands.php
mcsf: > My best guess here would be to add a completely new label to avoid breaking others, which aligns with the seemingly redundant strings you mentioned. Do you have any other recommendations?

Do you mean a new label which would be the same as `add_new_label` but with different casing expectations?

priethor: I originally meant this, but that's before realizing core `add_new_label` values have been updated in 6.8 to remove the `New` part, so what we should do is change the default value for `add_new_value` to align with core when creating CPTs in SCF (see #193).

PR #202, opened 2025-07-31
Project Management: Add zip creation script.

cbravobernal:
Use a zip creation script to avoid manual human errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,6 @@ assets/build/**/*.php
 /blob-report/
 /playwright/.cache/
 /artifacts/
+
+# Release folder
+release/
--- a/bin/create-release-zip.sh
+++ b/bin/create-release-zip.sh
@@ -1,0 +1,123 @@
+#!/bin/bash
+
+# Script to create a release zip file for Secure Custom Fields
+# This script creates a zip file with only the necessary files for distribution
+
+# Set script directory and define paths
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$SCRIPT_DIR"
+OUTPUT_DIR="$PROJECT_ROOT/release"
+ZIP_NAME="secure-custom-fields.zip"
+
+# Create output directory if it doesn't exist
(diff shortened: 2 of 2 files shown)


mcsf (commented):
* Is there anything we can [learn from Gutenberg](https://github.com/WordPress/gutenberg/blob/trunk/bin/build-plugin-zip.sh)? In particular, I'm wondering about measures to check the cleanliness of the local repo. I'm not suggesting adopting such a detailed script (with its diffing and user prompting), but a blunter approach of failing early if the Git stage isn't clean.

* Why not `set -e`? I see some error handling occasionally but not consistently. Many things can fail, including `mktemp`, `zip`, etc. Most of the current error handling is also yielding warnings rather than errors, and I'm not sure that's right.

* Unlike Gutenberg, where it's encouraged that volunteers of varying experience can help with the release, in SCF we can expect a handful of experienced plugin maintainers to be the audience of this script. As such, I think the verbose logging can be a distraction, and I'm more convinced that `set -e` is a better fit, letting the script abort if **any** command fails. Combine with `set -x` or sparse `echo` so that the maintainer running the script can quickly identify the issue if there is one.

* This script formalises where source files should live, but does nothing to detect stray files. It's too easy for a developer to add a "core file" (i.e. one at the root) during development and forget to update this script. The fact that there is a bit of boilerplate makes it marginally more difficult to quickly spot what is zipped and not. There are a few possible approaches here, I think:
  * It can be something simple but effective like a set of assertions:
    * Assert that certain files and directories exist, then add them to the ZIP file: PHP code, bundled JS, vendor assets, etc.
    * Assert that certain files and directories exist, then ignore them: JS source files, configuration, etc.
    * Assert that no other files or directories exist (i.e. fail if anything remains).
  * It could also (but not exclusively) rely on Git, because certain files are meant to be tracked and others aren't. The Gutenberg script does some of this: it calls `git clean` with certain options, thereby deleting unexpected files, and makes some modifications on this clean state.

* I have some low-level notes (e.g. testing the integrity of the zip file rather than its existence, shell idioms...) but I don't want that to get in the way of the important.

PR #204, opened 2025-07-31
Bug fix: The release script was moved to the bin folder. The root folder must be updated.

cbravobernal:
The release script was moved to the bin folder. The root folder must be updated.


mcsf (commented):
What about letting `git rev-parse --show-toplevel` handle it unambiguously for us?

PR #206, opened 2025-08-04
Allow binding Date block to `date_picker` fields

ockham:
As of https://github.com/WordPress/gutenberg/pull/70585, the Date block can be used with Block Bindings.

This PR allows connecting it to SCF's `date_picker` field.

date-block-binding-scf

### DIFF
--- a/assets/src/js/bindings/block-editor.js
+++ b/assets/src/js/bindings/block-editor.js
@@ -39,6 +39,9 @@ const BLOCK_BINDINGS_CONFIG = {
 		linkTarget: [ 'text', 'checkbox', 'select' ],
 		rel: [ 'text', 'checkbox', 'select' ],
 	},
+	'core/post-date': {
+		datetime: [ 'date_picker' ],
+	},
 };
 
 /**


ockham:
Note that if the Date block is inserted and its `datetime` attribute is bound to an ACF `date_picker` field as seen in the animated gif in the PR description, it will currently show the wrong date on the frontend. The reason is that the `datetime` attribute is set to the current timestamp upon insertion (the reason for this is to distinguish the block from its deprecated version, which didn't have a `datetime` attribute, and would instead automatically "bind" to the containing post's publish date).

However, the Date block's `datetime` attribute isn't listed as bindable [yet](https://github.com/WordPress/wordpress-develop/pull/9299/files#diff-c569b1ecb11a007a563788a481710b51e0cb21cd1d1885490bf9c91777842f23R288). While the Date block itself contains [logic to provide backwards-compatibility](https://github.com/WordPress/gutenberg/blob/dff10d2f731a15446aadae4895e79f3013dcb980/packages/block-library/src/post-date/index.php#L21-L59), that code is currently only invoked if the attribute is not set at all.

We'll probably need to amend the latter logic to accommodate for cases where the attribute _is_ set, but to the "wrong" (i.e. fallback) value, and needs to be overridden by Block Bindings.

_Edit: Filed https://github.com/WordPress/gutenberg/pull/71047 to address this._

---

In the long run, we'll need to find a way to avoid requiring complex orchestrations like this for anyone who wants to extend Block Bindings (by adding more block attributes as bindable), including Gutenberg. [Here's a related Core Trac ticket.](https://core.trac.wordpress.org/ticket/62090)

PR #207, opened 2025-08-20
Backport: Update SCF to include all 6.4.3 fixes.

cbravobernal:
## What

Apply all changes done in 6.4.3, including:

– Unsafe HTML in field group labels is now correctly escaped for conditionally loaded field groups, resolving a JS execution vulnerability in the classic editor
– HTML is now escaped from field group labels when output in the ACF admin
– Bidirectional and Conditional Logic Select2 elements no longer render HTML in field labels or post titles
– The acf.escHtml function now uses the third party DOMPurify library to ensure all unsafe HTML is removed. A new esc_html_dompurify_config JS filter can be used to modify the default behaviour
– Post titles are now correctly escaped whenever they are output by ACF code. Thanks to Shogo Kumamaru of LAC Co., Ltd. for the responsible disclosure
– An admin notice is now displayed when version 3 of the Select2 library is used, as it has now been deprecated in favor of version 4

### DIFF
--- a/assets/src/js/_acf-condition-types.js
+++ b/assets/src/js/_acf-condition-types.js
@@ -71,7 +71,7 @@
 		const template = function ( selection ) {
 			return (
 				`<span class="acf-${ typeAttr }-select-name acf-conditional-select-name">` +
-				acf.escHtml( selection.text ) +
+				acf.strEscape( selection.text ) +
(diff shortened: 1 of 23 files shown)


priethor (approved):
Works as expected 🫡

PR #208, opened 2025-08-22
Backport: Update 6.5.0

cbravobernal:
## What

Backports all changes from 6.5.0.

Needs testing before merging.

## Changelog

**Green ticks have been tested manually by me.**

- Flexible Content layouts can now be renamed in the post editor, giving content editors better clarity when managing layouts ✅ 
- Flexible Content layouts can now be disabled, preventing them from rendering on the frontend without needing to delete their data  ✅ 
- Flexible Content layouts can now be collapsed and expanded in bulk for faster content editing  ✅ 
- Editing a Flexible Content layout now highlights the layout being edited, making it easier to identify  ✅ 
- The Date and Date Time Picker fields can now be configured to default to the current date  ✅ 
- Custom Icon Picker tabs now work correctly when used inside an ACF Block ✅ 
- Duplicating a Field Group no longer causes a fatal error when using Russian translations ✅ 
- ACF classes no longer use dynamic class properties, improving compatibility with PHP 8.2+  ✅ 
- Field group metabox collapse and expand buttons are no longer misaligned in the post editor  ✅ 
- HTML is now escaped from field validation errors and tooltips ✅


priethor (commented):
Retroactively tested this, works as expected 🚀

PR #211, opened 2025-08-30
Get the `formatted_value` from the original field value

Mr2P:
The `formatted_value` property of the $field['name'] . '_source' should come from the original value, not the REST value. Other plugins might change the REST value using the filter `acf/rest/format_value_for_rest`, which can cause the `formatted_value` to be incorrect.

### DIFF
--- a/includes/rest-api/class-acf-rest-api.php
+++ b/includes/rest-api/class-acf-rest-api.php
@@ -232,10 +232,10 @@ class ACF_Rest_Api {
 
 				// Format the field value according to the request params.
 				$format = $request->get_param( 'acf_format' ) ?: acf_get_setting( 'rest_api_format' );
-				$value  = acf_format_value_for_rest( $value, $post_id, $field, $format );
+				$rest_value  = acf_format_value_for_rest( $value, $post_id, $field, $format );
 
 				// We keep this one for backward compatibility with existing code that expects the field value to be.
-				$fields[ $field['name'] ]             = $value;
+				$fields[ $field['name'] ]             = $rest_value;
 				$fields[ $field['name'] . '_source' ] = array(
 					'label'           => $field['label'],
 					'type'            => $field['type'],
--- a/includes/rest-api/class-acf-rest-api.php
+++ b/includes/rest-api/class-acf-rest-api.php
@@ -231,8 +231,8 @@ class ACF_Rest_Api {
 				}
 
 				// Format the field value according to the request params.
-				$format = $request->get_param( 'acf_format' ) ?: acf_get_setting( 'rest_api_format' );
-				$rest_value  = acf_format_value_for_rest( $value, $post_id, $field, $format );
+				$format     = $request->get_param( 'acf_format' ) ?: acf_get_setting( 'rest_api_format' );
+				$rest_value = acf_format_value_for_rest( $value, $post_id, $field, $format );
 
 				// We keep this one for backward compatibility with existing code that expects the field value to be.
 				$fields[ $field['name'] ]             = $rest_value;
--- a/includes/rest-api/class-acf-rest-api.php
+++ b/includes/rest-api/class-acf-rest-api.php
@@ -231,7 +231,7 @@ class ACF_Rest_Api {
 				}
 
 				// Format the field value according to the request params.
-				$format     = $request->get_param( 'acf_format' ) ?: acf_get_setting( 'rest_api_format' );
+				$format     = $request->get_param( 'acf_format' ) ? $request->get_param( 'acf_format' ) : acf_get_setting( 'rest_api_format' );
 				$rest_value = acf_format_value_for_rest( $value, $post_id, $field, $format );
 
 				// We keep this one for backward compatibility with existing code that expects the field value to be.


includes/rest-api/class-acf-rest-api.php
```diff
@@ -231,11 +231,11 @@ public function load_fields( $object, $field_name, $request, $object_sub_type )
...
 
 				// Format the field value according to the request params.
-				$format = $request->get_param( 'acf_format' ) ?: acf_get_setting( 'rest_api_format' );
-				$value  = acf_format_value_for_rest( $value, $post_id, $field, $format );
+				$format     = $request->get_param( 'acf_format' ) ?: acf_get_setting( 'rest_api_format' );
+				$rest_value = acf_format_value_for_rest( $value, $post_id, $field, $format );
 
 				// We keep this one for backward compatibility with existing code that expects the field value to be.
-				$fields[ $field['name'] ]             = $value;
+				$fields[ $field['name'] ]             = $rest_value;
```
cbravobernal, line 238: Thanks for the contribution!

Why will this fix your fatal error on the plugin? The change, as far as I can see, is just changing a variable name.

Mr2P: Hi, thanks for approving the pull request.
Here is my explanation. 

The original code from SCF:

// Get raw value.
$value = acf_get_value( $post_id, $field );

// First format for rest API for backward compatibility.
$value  = acf_format_value_for_rest( $value, $post_id, $field, $format );

// We keep this one for backward compatibility with existing code that expects the field value to be.
$fields[ $field['name'] ]             = $value;
$fields[ $field['name'] . '_source' ] = array(
  'label'           => $field['label'],
  'type'            => $field['type'],
  // This line re-formats the already formatted REST value instead of the raw value, which causes issues.
  'formatted_value' => acf_format_value( $value, $post_id, $field ),
);

I fixed it by renaming the formatted REST value to $rest_value and leaving the original $value untouched. This way, when calling acf_format_value() to populate the formatted_value property of _source, it’s based on the raw value, not on the already formatted value:

// Format value for rest API for backward compatibility.
$rest_value  = acf_format_value_for_rest( $value, $post_id, $field, $format );

// We keep this one for backward compatibility with existing code that expects the field value to be.
$fields[ $field['name'] ]             = $rest_value;
$fields[ $field['name'] . '_source' ] = array(
  'label'           => $field['label'],
  'type'            => $field['type'],
  'formatted_value' => acf_format_value( $value, $post_id, $field ),
);

In my plugin (Meta Field Block), I used the 'acf/rest/format_value_for_rest' filter to change the format of the REST value to support displaying all field types (even Group, Repeater, Flexible, Relationship, etc). When SCF re-format some multi-value field types like select, it treats the altered REST value as multiple values, which is why the error occurs.

priethor: Hi @Mr2P, thanks again for engaging here 🙂 

After a thorough read of the full method, I agree that the new variable name is helping not to overwrite the existing `$value` variable, great! My only concern is that I am not sure about the side effects this effect can have for other users; would you mind adding some tests to ensure backwards compatibility?

Mr2P: Hi @priethor,

I’m not sure what needs to be tested for this fix. The root cause of the issue is that, starting with SCF 6.5.2, the wrong value has been assigned to the new formatted_value property.

It should be:

$raw_value = acf_get_value( $post_id, $field );
$fields[ $field['name'] . '_source' ] = array(
  'formatted_value' => acf_format_value( $raw_value, $post_id, $field ),
);

not:

$raw_value = acf_get_value( $post_id, $field );
$formatted_value_for_rest = acf_format_value_for_rest( $raw_value, $post_id, $field, $format );
$fields[ $field['name'] . '_source' ] = array(
  'formatted_value' => acf_format_value( $formatted_value_for_rest, $post_id, $field ),
); 

The value returned by acf_format_value_for_rest is filtered by:

apply_filters( 'acf/rest/format_value_for_rest', $value_formatted, $post_id, $field, $value, $format );

If any third-party plugin uses this filter to change the REST value, the formatted_value property may end up with the wrong value.

priethor: Gotcha! From a logical point of view, it makes sense to me to run the formatting on the raw value instead of the already-formatted-for-REST value, especially considering you can hook into the latter.

I'm going to go ahead and proceed, but I'd still love it if @cbravobernal of @ockham could verify that there are no side effects in block bindings, as the `formatted_value` property, even if introduced in 6.5.2, is used there.

Thanks for bearing with us, @Mr2P, and please keep the feedback and testing coming ❤️

priethor (approved):

priethor:
I've gone ahead and fixed the Code Standard error that appeared in this file, although not specifically related to this PR.

includes/rest-api/class-acf-rest-api.php
Mr2P: @priethor Thanks for the approving. I'll keep testing on my end.

cbravobernal: There are no side effects on block bindings. As far as I tested. Let's merge it.

cbravobernal (approved):

Issue #162, opened 2025-05-21
Create Schemas for the current implementation of file-based definitions


priethor:
As of SCF 6.4, there are no JSON schemas to help developers and LLMs create JSON definitions from scratch, so there should be schemas available for:
- [X] Custom Post Types - #221
- [x] Field Groups - #287
  - [x] Basic fields - text, email, number, password, url, range, textarea #303
  - [x] Content fields: image, gallery, file, wysiwyg, oembed #304
  - [x] Choice fields: select, checkbox, radio, button_group, true_false #305
  - [x] Relational fields: post_object, relationship, user, taxonomy, page_link, link #306
  - [x] Advanced fields: date_picker, time_picker, color_picker, google_map, icon_picker #307
  - [x] Layout fields: group, repeater, flexible_content, clone, tab, accordion, message #308
- [X] Taxonomies - #261
- [X] Options Pages - #270

PR #221, opened 2025-10-20
JSON Schemas: Add framework and Post Type schema

priethor:
## What
Part of #162.

Adds JSON schema validation infrastructure for post type definitions. This not only improves the creation of new Post Types via JSON files, but also makes integrating the Abilities API easier.

  ## Why
SCF supports exporting and importing its entities through JSON files, but the lack of schemas makes it difficult to create them from scratch; even LLMs have trouble creating a basic, working JSON file for post types, as seen in the included fixture of a wrong format.

Having these schemas will help create new post types from scratch by:
  - Clearly specifying what attibutes are accepted, required, etc.
  - Validating post type configurations before registration to prevent runtime errors
  - Providing clear validation feedback for invalid configurations, like when trying to use a WordPress reserved term
  - Enables better tooling support with schema definitions

  ## How
  - **Validator Class**: New `SCF_JSON_Schema_Validator` that accepts different format (JSON files, strings, or PHP objects) and serves as a base for post type schemas and future ones.
  - **Schema Files**: Comprehensive JSON schema for post types with WordPress and SCF-specific validation rules
  - **Common Definitions**: Shared schema components including WordPress reserved terms, which willl come in handy for the other entities.
  - **Test Coverage**:added multiple tests to validate both positive and negative cases, including:
    - A real SCF export
    - An invalid file generated by LLMs to prove why the schemas are important
  - **Export Format Support**: SCF currently exports some things a bit raw, like booleans being `1`. The import and schema accept both strings and boolean values. The same goes with single/multiple post types, the import accepts both.

  ## Testing Instructions
  1. Run `composer test:php tests/php/PostTypeSchemaTest.php --verbose` - all post type schema validation tests should pass
  2. Verify schema validation works with both valid and invalid post type configurations adding them to the fixtures folder
  3. Most importantly, test with exported post type JSON files to ensure import format compatibility:
  3.1 Create a post type
3.2 Go to SCF -> Tools, export the post type
3.3 Put it in `tests/php/fixtures/schemas/post-types/valid`
3.4 Run again the test via `composer test:php tests/php/PostTypeSchemaTest.php --verbose`, it should pass.


schemas/post-type.schema.json
```diff
@@ -0,0 +1,527 @@
...
+					"type": "string",
+					"pattern": "^[a-z0-9_-]+$",
+					"minLength": 1,
+					"maxLength": 20,
+					"not": {
+						"$ref": "file://common.schema.json#/definitions/wordpressReservedTerms"
+					},
+					"description": "The post type key passed to register_post_type() (e.g. 'book', 'product'). Max 20 characters, lowercase letters/numbers/underscores/dashes only, cannot be WordPress reserved terms."
+				},
+				"ID": {
```
cbravobernal: I exported a json post type and could not see any ID, I guess it will be internal.

cbravobernal (approved):
It's a long PR, I couldn't find anything weird or out of a default export.
I guess we may find some issues in a future when is used, but we need that testing / feedback.

schemas/post-type.schema.json
priethor: It is internal, but upon a more detailed check, I've noticed it's actually always ignored in favor of the `key` property, so:
- When importing a new Post Type, the ID is not used and is assigned upon DB insertion.
- When importing an existing Post Type, the ID is overwritten with the actual ID.

All in all, it's only metadata that represents the Post Type, but since it's never used during import or export, let's remove it.

Good catch! ☺️

PR #229, opened 2025-11-11
Backports: Updates 6.6.0, 6.6.1, 6.6.2

cbravobernal:
## What

Backports from the last versions.

Includes a new file structure for v3 blocks.


priethor (commented):
Shouldn't we add a GH workflow to run the new JS unit tests? Or make them directly e2e tests and avoid needing an extra layer of testing.

cbravobernal:
> Shouldn't we add a GH workflow to run the new JS unit tests? Or make them directly e2e tests and avoid needing an extra layer of testing.

I prefer to have some unit JS tests if possible, as they cannot be flaky and still test some functions. But is true we should run them on the GH workflow 😅

priethor:
> But is true we should run them on the GH workflow 😅

Added them in #250 ❤️ . Note that the tests added in this PR seem to fail 😅

PR #231, opened 2025-11-11
Continuous integration: add integration with Codecov

priethor:
### What
This PR improves the SCF continuous integration by integrating with [Codecov](https://about.codecov.io/).

### Why
As a fork, SCF currently has a low code coverage:
- We should ensure that new features do get coverage
- Ideally, as we add coverage to legacy code, it will help us ensure legacy code doesn't break and understand future backports

### How
By generating a PHP coverage report during the PHPUnit action, and uploading the results to Codecov


cbravobernal (approved):
Let's see. 🚀

PR #220, opened 2025-10-15
Documentation: Regenerate manifest

priethor:
Closes #218 

A few documentation pages were not being deployed because they were missing from the generated `manifest.json`. This PR regenerates the manifest to include all documentation pages, fixing the 404 errors.

The missing pages were:
- tutorials/first-custom-field
- tutorials/first-options-page
- tutorials/first-taxonomy
 - welcome/how-content-works

PR #232, opened 2025-11-11
GitHub Actions: Warn if docs manifest possibly outdated

mcsf:
Feel free to close if useless!

Noticing the problem in #220, I drew inspiration from [one of Gutenberg's workflows](https://github.com/WordPress/gutenberg/blob/trunk/.github/workflows/check-components-changelog.yml) to add a workflow that will fail if `docs/bin/manifest.json` needs to be regenerated.

It does this by running the `docs:manifest` composer script and verifying that the manifest has not changed. If it has, something is wrong with the branch, and the developer is told to manually generate the docs. I prefered the manual approach for several reasons, one of which is to avoid "magic" and instead teaching/reminding the developer of processes.

I attempted to fit into the style and practices of this repo's other workflows, but I may be wildly wrong. I'll also let you folks label this as you see fit. :)

### DIFF
--- a/.github/workflows/docs-manifest.yml
+++ b/.github/workflows/docs-manifest.yml
@@ -1,0 +1,39 @@
+name: Keep docs manifest up to date
+on:
+  pull_request:
+    types: [opened, synchronize]
+    branches:
+      - trunk
+    paths:
+      - docs/**/*.md
+
+jobs:
+  check-docs-manifest:
+    name: Check docs manifest for staleness
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Setup PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: '8.4'
+        tools: composer:v2
+
+    - name: Install dependencies
+      run: composer install
+
+    - name: Regenerate manifest
+      run: composer run docs:manifest
+
+    - name: Check whether manifest has been updated
+      env:
+        MANIFEST_PATH: docs/bin/manifest.json
+      run: |
+        if ! git diff --quiet -- "$MANIFEST_PATH"; then
+          echo "There are documentation changes in this branch that require that the docs manifest be regenerated."
+          echo
+          echo "Please run 'composer run docs:manifest' and commit $MANIFEST_PATH"
+          exit 1
+        fi
--- a/.github/workflows/docs-manifest.yml
+++ b/.github/workflows/docs-manifest.yml
@@ -1,11 +1,10 @@
 name: Keep docs manifest up to date
 on:
   pull_request:
-    types: [opened, synchronize]
     branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-manifest.yml
```diff
@@ -0,0 +1,40 @@
+name: Keep docs manifest up to date
+on:
+  pull_request:
+    types: [opened, synchronize]
+    branches:
+      - trunk
+    paths:
+      - docs/**/*.md
```
mcsf: I made an assumption with this `path` param in an attempt to save resources.

.github/workflows/docs-manifest.yml
```diff
@@ -0,0 +1,40 @@
+name: Keep docs manifest up to date
+on:
```
mcsf, line 2: The `on` config here is what diverges the most from the other workflows. Initially, based on Gutenberg's components-changelog workflow, I wanted this check to be more conservative.

But now that I look at it again, and given the approach I finally took, maybe it should be the same as in the others. Thoughts?

.github/workflows/docs-manifest.yml
```diff
@@ -0,0 +1,39 @@
...
+    - uses: actions/checkout@v4
+
+    - name: Setup PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: '8.4'
+        tools: composer:v2
+
+    - name: Install dependencies
+      run: composer install
```
priethor: Can we use `--no-dev` to make it faster? We could even just run the PHP script directly to save resources; it has no dependencies.

.github/workflows/docs-manifest.yml
priethor: Nice one! Maybe worth adding `docs/bin/**` too for PRs that affect the docs tooling.

.github/workflows/docs-manifest.yml
priethor: Looks good to me like this. What do you think could be different?

priethor (approved):
Love to see this!

.github/workflows/docs-manifest.yml
mcsf: I wasn't sure about the preferences — if any — for the `branches` property, nor about which workflows should also run on `push` events. But I'm about to push a simplification here, limiting the workflow to `pull_request` events and looking more like other workflows.

.github/workflows/docs-manifest.yml
mcsf: Good point! I'll remove it altogether.

.github/workflows/docs-manifest.yml
priethor: Pull request events sound good, regenerating the manifest should be enforces for PRs that target `trunk` 👍

PR #236, opened 2025-11-12
[Testing CI] dummy docs modification

mcsf:
Testing #232


mcsf:
Nice, it works. I first pushed a change to the contents of a docs file, which triggered the workflow but caused no failure, as expected. I then pushed a file rename, which definitely requires a new manifest, and the workflow correctly failed:

Screenshot 2025-11-12 at 16 51 17
Screenshot 2025-11-12 at 16 51 25

The work here is done. Closing.

PR #239, opened 2025-11-14
Bugfix: ensure the update post type ability doesn't create new posts instead

priethor:
## What
Fixes the update post ability to require the ID, preventing the creation of new posts instead.

## Why
Previously, the ability required either an ID, a key, or a slug. However, the internal `acf_update_post_type` creates a new post if not provided with the ID it needs to update, so updates could result in new posts.

## How
- Schema change: Modified `register_update_post_type_ability()` to use `get_post_type_with_internal_fields_schema()` with only `ID` marked as required, making all other fields optional.
- Callback enhancement: Updated `update_post_type_callback()` to merge incoming input with existing post type data, ensuring that only specified fields are updated while preserving original values for the rest.

### DIFF
--- a/includes/abilities/class-scf-post-type-abilities.php
+++ b/includes/abilities/class-scf-post-type-abilities.php
@@ -274,6 +274,11 @@ class SCF_Post_Type_Abilities {
 	 * @since 6.6.0
 	 */
 	private function register_update_post_type_ability() {
+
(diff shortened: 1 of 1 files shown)


priethor:
I'll go ahead and merge this despite the codecov flag because I still need to push the e2e tests for abilities _and_ configure codecov uploads from e2e tests.

PR #240, opened 2025-11-14
Bugfix: Use specific entity prefixes for key generation when duplicating entities

priethor:
## What
Fixes a bug where duplicated entities would always have the `group_` prefix instead of the prefix of the entity.

## Why
It's inconsistent with the keys of new entities, and won't pass schema validations.

## How
In the base class `ACF_Internal_Post_Type`, use the attribute `post_key_prefix`, which child classes override, instead of hardcoding the prefix.

## Testing instructions
1. Create a post type
2. Duplicate it
3. Check that its internal key doesn't start with `group_` anymore, but with `post_type_`

### DIFF
--- a/includes/class-acf-internal-post-type.php
+++ b/includes/class-acf-internal-post-type.php
@@ -724,7 +724,7 @@ if ( ! class_exists( 'ACF_Internal_Post_Type' ) ) {
 
 			// Update attributes.
 			$post['ID']  = $new_post_id;
-			$post['key'] = uniqid( 'group_' );
+			$post['key'] = uniqid( $this->post_key_prefix );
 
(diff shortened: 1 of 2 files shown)


includes/class-acf-internal-post-type.php
```diff
@@ -724,7 +724,7 @@ public function duplicate_post( $id = 0, $new_post_id = 0 ) {
 
 			// Update attributes.
 			$post['ID']  = $new_post_id;
-			$post['key'] = uniqid( 'group_' );
+			$post['key'] = uniqid( $this->post_key_prefix );
```
cbravobernal: Let's add a comment to differentiate from future backports.

PR #241, opened 2025-11-14
Blocks V3: Fix flexible content not working in sidebar - modal.

cbravobernal:
## What

The new V3 Blocks allow to edit the content in the sidebar and in a modal. Flexible content is not working, if you set it as required.

https://github.com/user-attachments/assets/3427bc81-eda6-4976-ac71-d73e784f36b3

## Fix

To fix it, it needs to:
 - Skip server validation if is on a block - sidebar. Although block data is still validated, sanitized, with nonce and capability checks.
 - Serialize flexible content data to be validated on the editor.

## Screenshare

https://github.com/user-attachments/assets/ec60711b-e83b-415f-a091-1c7cc4eda746

### DIFF
--- a/assets/src/js/_acf.js
+++ b/assets/src/js/_acf.js
@@ -882,7 +882,6 @@
 	 * Convert an object with numeric string keys to a sorted array.
 	 * Example: {"0": "one", "2": "three", "1": "two"} becomes ["one", "two", "three"]
 	 *
-	 * @date    14/01/25
 	 * @since   SCF 6.6.0
 	 *
 	 * @param   object obj The object to convert
--- a/assets/src/js/_acf.js
+++ b/assets/src/js/_acf.js
@@ -860,17 +860,18 @@
 	 * Used to detect objects that should be converted to arrays (e.g., checkbox values).
 	 *
 	 * @since   SCF 6.6.0
+	 * @private
 	 *
 	 * @param   object obj The object to check
 	 * @return  boolean True if all keys are numeric strings
 	 */
-	acf.hasOnlyNumericKeys = function ( obj ) {
-		var keys = Object.keys( obj );
+	const hasOnlyNumericKeys = function ( obj ) {
+		const keys = Object.keys( obj );
 		if ( keys.length === 0 ) {
 			return false;
 		}
 
-		for ( var i = 0; i < keys.length; i++ ) {
+		for ( let i = 0; i < keys.length; i++ ) {
 			if ( ! /^\d+$/.test( keys[ i ] ) ) {
 				return false;
 			}
@@ -883,13 +884,14 @@
 	 * Example: {"0": "one", "2": "three", "1": "two"} becomes ["one", "two", "three"]
 	 *
 	 * @since   SCF 6.6.0
+	 * @private
 	 *
 	 * @param   object obj The object to convert
 	 * @return  array The sorted array
 	 */
-	acf.numericObjectToArray = function ( obj ) {
-		var arr = [];
-		var keys = Object.keys( obj )
+	const numericObjectToArray = function ( obj ) {
+		const arr = [];
+		const keys = Object.keys( obj )
 			.map( function ( k ) {
 				return parseInt( k, 10 );
 			} )
@@ -897,7 +899,7 @@
 				return a - b;
 			} );
 
-		for ( var i = 0; i < keys.length; i++ ) {
+		for ( let i = 0; i < keys.length; i++ ) {
 			arr.push( obj[ keys[ i ].toString() ] );
 		}
 		return arr;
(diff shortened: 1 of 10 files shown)


priethor (commented):
Can you rebase/merge and update your branch? I've just lowered Codecov's threshold for new patches in https://github.com/WordPress/secure-custom-fields/pull/242

cbravobernal:
> Can you rebase/merge and update your branch? I've just lowered Codecov's threshold for new patches in #242

Done.

priethor:
Strange, the Codecov comment is not getting updated 🤔

cbravobernal:
It says I'm ok: 

[codecov/patch](https://github.com/WordPress/secure-custom-fields/pull/241/checks?check_run_id=55421575968)
codecov/patch — 80.00% of diff hit (target 1.17%)

priethor:
Now it updated, but didn't pick the threshold properly 🫠

cbravobernal:
> Now it updated, but didn't pick the threshold properly 🫠

I think I'm going to merge it, anyway 😆

assets/src/js/_acf.js
```diff
@@ -855,6 +855,148 @@
...
+	/**
+	 * Check if an object has only numeric string keys.
+	 * Used to detect objects that should be converted to arrays (e.g., checkbox values).
+	 *
+	 * @since   SCF 6.6.0
+	 *
+	 * @param   object obj The object to check
+	 * @return  boolean True if all keys are numeric strings
+	 */
+	acf.hasOnlyNumericKeys = function ( obj ) {
```
priethor: How should this handle edge cases like `00123` or negative values? Maybe that can't happen in the functionality currently using this function, but for such a global function, it would be good to clarify it.

It's probably something `acf` itself shouldn't handle as a public API.

includes/forms/form-post.php
```diff
@@ -83,6 +83,36 @@ public function initialize() {
 			add_action( 'add_meta_boxes', array( $this, 'add_meta_boxes' ), 10, 2 );
 		}
 
+		/**
+		 * Checks if a field group is assigned to blocks.
+		 *
+		 * Block field groups should not be rendered as metaboxes because:
+		 * 1. They are managed by the block editor (blocks v3)
+		 * 2. They have their own validation system via AJAX
+		 * 3. Rendering them as metaboxes causes duplicate validation on post save
```
priethor, line 92: The PR description says we skip validation in these cases; could we add tests to ensure validation is still performed by other means?

priethor (commented):
>  think I'm going to merge it, anyway 😆

Indeed that's not a blocker, but I left more important feedback 🙂

assets/src/js/_acf.js
cbravobernal: Updated.

includes/forms/form-post.php
cbravobernal: Updated.

tests/e2e/block-testimonial.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+		);
+		await roleField.fill( 'CEO, Example Company' );
+		// Blur to trigger update
+		await roleField.blur();
+
+		// Wait for block preview to update
+		await page.waitForTimeout( 1000 );
+
+		// Get the editor canvas (content is in an iframe in newer WordPress versions)
+		const canvas = await editor.canvas;
```
priethor, line 83: I think this is ok since we are only running e2e tests on the latest WP version, even if we support older ones. Should we consider running them in older versions?

tests/e2e/plugins/scf-test-testimonial-block.php
```diff
@@ -0,0 +1,84 @@
...
+add_action( 'init', 'scf_test_register_testimonial_block' );
+
+/**
+ * Register field group for the testimonial block.
+ */
+function scf_test_register_testimonial_fields() {
+	// Check if ACF function exists.
+	if ( ! function_exists( 'acf_add_local_field_group' ) ) {
+		return;
+	}
```
priethor, line 32: When would this function not exist during tests?

assets/src/js/_acf.js
priethor: Thanks for making it private 👍

priethor (approved):

tests/e2e/block-testimonial.spec.ts
cbravobernal: I wouldn't do it right now. Classic editor should be soon deprecated by default.

tests/e2e/plugins/scf-test-testimonial-block.php
cbravobernal: It wouldn't 😓  But it may prevent fatals if they bad-deprecate it.

PR #245, opened 2025-11-18
Fix: Validate required JSON schemas before registering abilities 

priethor:
## What
Adds validation check to `SCF_Post_Type_Abilities` that ensures all required schema files exist before registering WordPress Abilities API hooks.

## Why
Without this, missing schemas cause abilitiies registration to crash at runtime. Now, if schemas are missing, abilities simply don't register and will display an admin notice

## How
- Added schema validation in `__construct()` that checks `validate_required_schemas()` before registering hooks
- Changed from `require_once` + `new` to `acf_get_instance()` for validator initialization
- Early return if validation fails (abilities won't register)

## Testing Instructions
1. Run `composer test:php` - all 58 tests pass
2. Verify abilities register normally when all schema files present
3. Verify abilities don't register if schema files missing

### DIFF
--- a/includes/abilities/class-scf-post-type-abilities.php
+++ b/includes/abilities/class-scf-post-type-abilities.php
@@ -13,606 +13,614 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
-// Include the validator class for static analysis.
-if ( ! class_exists( 'SCF_JSON_Schema_Validator' ) ) {
-	require_once ACF_PATH . 'includes/class-scf-json-schema-validator.php';
-}
-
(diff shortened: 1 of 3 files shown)


priethor:
Note that the coverage report is complaining about the `class-scf-post-type-abilities.php` file; most lines have changed because of nested indentation, so codecov considers we edited them as part of this patch.

I have an incoming PR just to test the Post Type abilities class and shouldn't be a blocker here.

cbravobernal (approved):
LGTM. 
I removed the schema files, the tests failed as expected and no WSOD on the admin.

PR #247, opened 2025-11-19
Codecov: Remove `coverage.changes` section from codecov settings

priethor:
## What
Removes the `changes` section from codecov settings

## Why
Codecov's YAML validator was complaining about the target options for this setting. This patch removes the setting entirely, as it's only for anomalies outside the patch in question.

## Testing
When the Codecov comment appears in this PR, the link for this PR's report within Codecov shouldn't show the same error as with the rest of PRs:
image

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -25,12 +25,6 @@ coverage:
                 threshold: 5%
                 if_ci_failed: ignore
 
-        changes:
-            default:
-                target: 70%
-                threshold: 5%
-                if_ci_failed: ignore
-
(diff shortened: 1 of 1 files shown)


priethor:
Confirming that the error doesn't appear anymore in [the report](https://app.codecov.io/gh/WordPress/secure-custom-fields/pull/247?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=WordPress).

PR #249, opened 2025-11-20
Translations: Update pot file for .org translations.

cbravobernal:
## What

We need to update the translations for the plugin. Since a couple of versions, we added some new strings and many changed their place on the code.

Question: can we commit this without making a new release?


kraftbj:
Noting that this would be helpful for people translating outside of translate.wordpress.org. It isn't needed for WP language packs.

cbravobernal:
> Noting that this would be helpful for people translating outside of translate.wordpress.org. It isn't needed for WP language packs.

Thanks for the info, let's close the PR then.

PR #255, opened 2025-11-27
Post Type Abilities: Return 404 if not found when deleting or duplicating

priethor:
## What
Ensures all SCF post type abilities return proper HTTP 404 errors when a post type is not found.

## Why
- Delete and duplicate abilities were not checking if the post type existed before operating, leading to generic failure errors instead of proper "not found" responses
- Error responses were inconsistent - some returned HTTP status codes, others didn't

## How
- Added existence checks to `delete_post_type_callback` and `duplicate_post_type_callback` before performing operations
- Created a private `post_type_not_found_error()` helper that returns `WP_Error` with 404 HTTP status
- Unified all "not found" error responses across get, update, delete, duplicate, and export abilities

## Testing Instructions
1. Call `scf/delete-post-type` with a non-existent identifier
2. Response should return error code `post_type_not_found` with HTTP status 404
3. Same for `scf/duplicate-post-type`, `scf/get-post-type`, `scf/update-post-type`, and `scf/export-post-type`

### DIFF
--- a/includes/abilities/class-scf-post-type-abilities.php
+++ b/includes/abilities/class-scf-post-type-abilities.php
@@ -358,7 +358,7 @@ if ( ! class_exists( 'SCF_Post_Type_Abilities' ) ) :
 				'scf/duplicate-post-type',
 				array(
 					'label'               => __( 'Duplicate Post Type', 'secure-custom-fields' ),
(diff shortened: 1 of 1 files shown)


priethor:
> ❌ Patch coverage is 0% with 15 lines

Full suite of e2e tests for abilities in a follow-up PR

PR #256, opened 2025-11-27
Post Type Abilities: Add e2e tests

priethor:
## What
Add comprehensive e2e tests for SCF Post Type Abilities.

## Why
Ensures the Post Type abilities work correctly through the WordPress Abilities API

## How

- Added new test file `tests/e2e/abilities-post-types.spec.ts`
- Tests all 8 post type abilities
- Tests call REST API endpoints directly (no UI interaction or MCP)
- Each test includes success cases and error handling (404 not found, 400 invalid input)
- Tests skip automatically on WordPress versions without the Abilities API

## Testing Instructions

1. Run `npm run test:e2e tests/e2e/abilities-post-types.spec.ts`
2. All tests should pass on WordPress 6.9+ or with the Abilities API plugin installed
3. Tests should skip gracefully on older WordPress versions or without the plugin

PR #257, opened 2025-11-27
Continuous Integration: Add WP version matrix to test PRs against WordPress trunk

priethor:
## What
Add WP version matrix to CI to test PRs against trunk, and fix failing tests against trunk.

## Why
- In general, testing against trunk helps catch compatibility issues early, but in particular, to test the Abilities only available in the unreleased 6.9 version
- E2e tests fail on WP trunk because the Meta Boxes panel is now collapsed by default

## How
- Add WP version matrix (`latest` + `trunk`) to e2e workflow
- On PRs, only test WP trunk until WP 6.9 releases with the Abilities API, when we can revert back to latest
- Fix existing tests by expanding the Meta Boxes panel if collapsed before interacting with SCF fields (workaround for [WordPress/gutenberg#72185](https://github.com/WordPress/gutenberg/issues/72185))

## Testing Instructions
1. Run e2e tests locally with WP trunk: `WP_ENV_CORE=WordPress/WordPress npm run test:e2e`
2. Tests should pass without "element is not visible" errors

### DIFF
--- a/tests/e2e/field-type-repeater.spec.ts
+++ b/tests/e2e/field-type-repeater.spec.ts
@@ -109,6 +109,17 @@ test.describe( 'Field Type > Repeater', () => {
         // Navigate to edit post page
         await admin.editPost( post.id );
 
+        // Wait for meta boxes to load and expand the panel if collapsed.
+        // @see https://github.com/WordPress/gutenberg/issues/72185
+        await page.waitForSelector( '.acf-postbox', { state: 'attached' } );
+        await page.evaluate( () => {
+            document.querySelectorAll( 'button' ).forEach( ( btn ) => {
+                if ( btn.textContent?.includes( 'Meta Boxes' ) && btn.getAttribute( 'aria-expanded' ) === 'false' ) {
+                    btn.dispatchEvent( new MouseEvent( 'click', { bubbles: true } ) );
+                }
+            } );
+        } );
+
         // Add a repeater row - using the correct selector matching the actual DOM element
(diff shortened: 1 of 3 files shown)


priethor:
Merging this as a blocker of #256, but feel free to add a post-merge review, @cbravobernal

tests/e2e/field-type-repeater.spec.ts
```diff
@@ -109,6 +109,17 @@ test.describe( 'Field Type > Repeater', () => {
...
         await admin.editPost( post.id );
 
+        // Wait for meta boxes to load and expand the panel if collapsed.
+        // @see https://github.com/WordPress/gutenberg/issues/72185
+        await page.waitForSelector( '.acf-postbox', { state: 'attached' } );
+        await page.evaluate( () => {
+            document.querySelectorAll( 'button' ).forEach( ( btn ) => {
+                if ( btn.textContent?.includes( 'Meta Boxes' ) && btn.getAttribute( 'aria-expanded' ) === 'false' ) {
+                    btn.dispatchEvent( new MouseEvent( 'click', { bubbles: true } ) );
+                }
```
cbravobernal, line 119: Wow, is not common to see this in e2e testing. You are opening all meta boxes by default here?

priethor: It's only opening the meta box panel at the bottom, but _you are absolutely right_ we shouldn't be using synthetic events. Refactored to use standard Playwright methods in #259

PR #259, opened 2025-11-27
e2e tests: refactor meta box panel expansion to use Playwright methods

priethor:
Follow up to https://github.com/WordPress/secure-custom-fields/pull/257#discussion_r2569471649

Refactors meta box panel expansion in e2e tests to use standard Playwright methods instead of page.evaluate()


cbravobernal (approved):
LGTM

PR #267, opened 2025-12-02
e2e: Fix broken tests in trunk against WP versions under 6.9

priethor:
## What
Fixes e2e tests that are broken in trunk after #259 

## Why
The refactor introduced in #259 was incompatible with WP 6.8, but I didn't catch the test regression as the PR was running the tests against trunk, not 6.8.

## How
Only trying to open the metabox panel if the button to open it exists, which will support both cases where the metaboxes panel is open and collapsed.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -35,7 +35,7 @@ jobs:
                     - event: 'pull_request'
                       node: '24'
                     - event: 'pull_request'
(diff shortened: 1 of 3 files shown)


priethor:
Confirming the fix works with WP `latest` (6.8 as of now), which was failing in trunk. Reverting the e2e testing matrix.

PR #268, opened 2025-12-03
Prepare 6.7.0 Release

cbravobernal:
*Release Date 2 Dec 2025*

*Features*

- Tested compatibility up to WordPress 6.9.
- Abilities support. Taxonomy abilities.
- JSON schemas. Taxonomy schema.


docs/bin/manifest.json
```diff
@@ -29,11 +29,6 @@
         "parent": null,
         "markdown_source": "https://github.com/wordpress/secure-custom-fields/blob/trunk/docs/welcome/index.md"
     },
-    "code-reference/abilities": {
-        "slug": "abilities",
-        "parent": "code-reference",
-        "markdown_source": "https://github.com/wordpress/secure-custom-fields/blob/trunk/docs/code-reference/abilities/index.md"
-    },
```
cbravobernal, line 36: @priethor can you please check why is removing the abilities code reference on docs?

Could it be due to refactors?

priethor: It seems related to the refactor, but there should at least be a docs reference for the base class. Is that the output generated by `composer run docs:manifest`?

priethor (approved):
I don't think the doc reference for an abstract class should be a blocker for this update; let's follow up after the release.

docs/bin/manifest.json
cbravobernal: Yes. I didn't change anything manually :)

PR #271, opened 2025-12-03
e2e tests: fix tests compatibility for WP 6.2

priethor:
## What
Fixes the e2e suite so that it also works with WordPress 6.2

## Why
SCF requires a minimum WordPress version of 6.2, but the wp test utils we are using don't support older versions 

## How
- Creating an extended `editor` object that provides compatible ways to get:
  - The preview page (the button text changed in WP 6.3)
  - The canvas (there was no iframe in 6.2)
- Switching term/user tests from ID selectors to text locators

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -30,11 +30,12 @@ jobs:
                 node: ['22', '24']
                 wp: ['6.2', 'latest', 'trunk']
                 exclude:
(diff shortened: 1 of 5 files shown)


priethor:
Confirming [tests pass in CI against WP 6.2](https://github.com/WordPress/secure-custom-fields/actions/runs/19954089691/job/57219712127?pr=271), reverting to use `latest`.

PR #274, opened 2025-12-05
JSON Schemas: fix `menu_icon` property in Options Pages schema

priethor:
## What
Fixes a bug in the `menu_icon` property that came up while adding Options Pages abilities in #273, as it had the wrong type.


cbravobernal:
I'm afraid this didn't make it to 6.7.1

PR #275, opened 2025-12-09
JSON Schemas: fix too-early validation of schemas in FTP hosts

priethor:
## What
Fix fatal error `Call to undefined function wp_generate_password()` on specific hosts, as reported in the forums: https://wordpress.org/support/topic/fatal-error-on-plugin-update-to-6-6-0/

## Why
Trying to be extra defensive, the JSON Schema validator validated the existence of the schemas during its construction. On hosts where WP uses FTP/SSH filesystem access, the call to`WP_Filesystem()` to load the schemas triggers code that calls `wp_generate_password()` - a pluggable function not yet available during early plugin load:
- The JSON schema validator was calling `WP_Filesystem()` in the constructor, before WordPress was fully initialized
- `WP_Filesystem` is overkill for reading local plugin-bundled JSON files (it was used to satisfy PHPStan requirements).

## How
- Removing the constructor schema validation (schemas will now be fully lazy loaded)
- Replacing `WP_Filesystem` with native `file_get_contents()` for reading local schema files
	- Plus adding PHPCS global exclusion for `file_get_contents`

## Testing Instructions
1. Update the plugin and verify no fatal error on activation
2. Use the abilities API and verify the schema validation still works
3. Run `vendor/bin/phpunit --filter "Test_SCF_JSON_Schema_Validator"`

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -24,6 +24,8 @@
 		<!-- Refactoring of this scale is not in scope yet.-->
 		<exclude name="WordPress.Files.FileName" />
 		<!-- Exclude the entire filename rule -->
+		<exclude name="WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents" />
+		<!-- file_get_contents is safe for reading local plugin files bundled with the plugin. -->
 	</rule>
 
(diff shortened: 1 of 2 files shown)


cbravobernal (approved):
LGTM

PR #281, opened 2025-12-11
Backports: Updates from 6.7.0

cbravobernal:
## What

Backports from 6.7.0.

As uncompiling 6.7.0.2 with inline editing is being harder than expected. And not working 100% like it should. This version puts the compiled version into pro blocks in order to work as the main source.

I'm thinking about a better approach, less React intense based for inline editing that I will experiment in future PRs, but we cannot stall this version much longer.

### DIFF
--- a/assets/src/js/_acf-tinymce.js
+++ b/assets/src/js/_acf-tinymce.js
@@ -111,7 +111,6 @@
 					if ( ! $textarea.closest( '.attachment-info' ).length ) {
 						$textarea.trigger( 'change' );
 					}
-					$textarea.trigger( 'change' );
 				} );
 
 				ed.on( 'blur', function ( e ) {
--- a/assets/src/js/_acf-tinymce.js
+++ b/assets/src/js/_acf-tinymce.js
@@ -108,12 +108,22 @@
 			init.setup = function ( ed ) {
 				ed.on( 'change', function ( e ) {
 					ed.save(); // save to textarea
+					if ( ! $textarea.closest( '.attachment-info' ).length ) {
+						$textarea.trigger( 'change' );
+					}
 					$textarea.trigger( 'change' );
+				} );
+
+				ed.on( 'blur', function ( e ) {
+					if ( $textarea.closest( '.attachment-info' ).length ) {
+						ed.save();
+						$textarea.trigger( 'change' );
+					}
 				} );
 
 				// Fix bug where Gutenberg does not hear "mouseup" event and tries to select multiple blocks.
 				ed.on( 'mouseup', function ( e ) {
-					var event = new MouseEvent( 'mouseup' );
+					const event = new MouseEvent( 'mouseup' );
 					window.dispatchEvent( event );
 				} );
 
--- a/assets/src/js/_acf-validation.js
+++ b/assets/src/js/_acf-validation.js
@@ -1177,9 +1177,10 @@
 										new CustomEvent(
 											'acf/block/has-error',
 											{
-												acfBlocksWithValidationErrors: [
-													block,
-												],
+												detail: {
+													acfBlocksWithValidationErrors:
+														block.clientId,
+												},
 											}
 										)
 									);
--- a/assets/src/js/pro/blocks-v3/components/block-edit.js
+++ b/assets/src/js/pro/blocks-v3/components/block-edit.js
@@ -14,23 +14,18 @@ import {
 } from '@wordpress/element';
 
 import {
-	BlockControls,
 	InspectorControls,
 	useBlockProps,
 	useBlockEditContext,
 } from '@wordpress/block-editor';
-import {
-	Button,
-	ToolbarGroup,
-	ToolbarButton,
-	Placeholder,
-	Spinner,
-	Modal,
-} from '@wordpress/components';
+import { Button, Placeholder, Spinner, Modal } from '@wordpress/components';
 import { BlockPlaceholder } from './block-placeholder';
 import { BlockForm } from './block-form';
 import { BlockPreview } from './block-preview';
 import { ErrorBoundary, BlockPreviewErrorFallback } from './error-boundary';
+import { BlockToolbarFields } from './block-toolbar-fields';
+import { InlineEditingToolbar } from './inline-editing-toolbar';
+import { PopoverWrapper } from './popover-wrapper';
 import {
 	lockPostSaving,
 	unlockPostSaving,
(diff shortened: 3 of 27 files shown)


jacobodonnell:
After this is fully backported to SCF, it seems like it the plugin will be at a really good point to fully diverge from ACF. By that, I mean the baseline namespace should be converted to `scf` instead of `acf`. Obviously there would have to be wrapper functions using the old `acf` namespace so it is easy to switch from ACF Pro to SCF. 

With inline editing support for blocks, SCF will be basically feature complete at that point. Any new features that come about would abstractions over new baseline Gutenberg/FSE features, so they can be built from the ground up instead of back ported from whatever the ACF devs comes up with. I don't there are any major features on the ACF roadmap that would justify further backporting.

cbravobernal:
> After this is fully backported to SCF, it seems like it the plugin will be at a really good point to fully diverge from ACF. By that, I mean the baseline namespace should be converted to `scf` instead of `acf`. Obviously there would have to be wrapper functions using the old `acf` namespace so it is easy to switch from ACF Pro to SCF.
> 
> With inline editing support for blocks, SCF will be basically feature complete at that point. Any new features that come about would abstractions over new baseline Gutenberg/FSE features, so they can be built from the ground up instead of back ported from whatever the ACF devs comes up with. I don't there are any major features on the ACF roadmap that would justify further backporting.

Thanks for the comment. Is not easy to determine which would be the best stop to fully diverge. Inline editing is a huge feature that I couldn't 1:1 with only the compiled version, and I'm running out of time before a long break ( 3 months )

Regarding new features, we are a community plugin, so we can be closer to Core in terms of using the new `Dataviews` or `Dataform` packages. O adapt to the new `contentOnly` mode done in Gutenberg.

Adapting all these custom fields to the Core way could be a huge win.

assets/src/js/_acf-tinymce.js
```diff
@@ -108,12 +108,22 @@
 			init.setup = function ( ed ) {
 				ed.on( 'change', function ( e ) {
 					ed.save(); // save to textarea
+					if ( ! $textarea.closest( '.attachment-info' ).length ) {
+						$textarea.trigger( 'change' );
+					}
 					$textarea.trigger( 'change' );
```
priethor: Can we double-check this? Is the possible double trigger intentional, or should it only be now inside the `if`?

tests/e2e/block-testimonial.spec.ts
```diff
@@ -158,20 +158,9 @@ test.describe( 'SCF Block > Testimonial', () => {
...
+		// Wait for the error notice to appear at the top of the editor
+		const errorNotice = page.locator( '.components-notice.is-error .components-notice__content' );
+		await expect( errorNotice ).toBeVisible( { timeout: 1000 } );
 
-		// Check for the error message within the field
-		const errorMessage = page.locator(
-			'.acf-field[data-name="quote"] .acf-notice.-error'
-		);
-		await expect( errorMessage ).toBeVisible();
-		await expect( errorMessage ).toContainText( 'required' );
```
priethor, line 175: Just to confirm, do the old assertions no longer apply, or is the test more lenient now?

cbravobernal: The old assertions no longer apply, as they now use default WP notifications.

assets/src/js/_acf-tinymce.js
cbravobernal: Nice catched!

includes/blocks.php
```diff
@@ -913,6 +940,22 @@ function ( $block ) {
...
+ * Enqueues scripts and styles to load inside the block editor iframe.
+ * This allows us to do things like style contenteditable, and other inline editing elements.
+ *
+ * @since ACF 6.7
+ */
+function acf_enqueue_in_iframe_styles() {
+	if ( is_admin() ) {
+		$min      = defined( 'ACF_DEVELOPMENT_MODE' ) && ACF_DEVELOPMENT_MODE ? '' : '.min';
+		$css_path = acf_get_url( "assets/build/css/pro/acf-styles-in-iframe-for-blocks{$min}.css" );
+		wp_enqueue_style( 'acf-inline-editing-styles', $css_path, ACF_VERSION, true );
```
priethor: Even if ported from upstream, this enqueuing doesn't seem to follow [the function signature](https://developer.wordpress.org/reference/functions/wp_enqueue_style/); it's passing `ACF_VERSION` as dependency and `true` as version. I'm not sure what `true` is meant to be here, but the call should probably be:

```suggestion
		wp_enqueue_style( 'acf-inline-editing-styles', $css_path, array(), ACF_VERSION );
```

assets/src/js/_acf-validation.js
```diff
@@ -1177,9 +1177,10 @@
 										new CustomEvent(
 											'acf/block/has-error',
 											{
-												acfBlocksWithValidationErrors: [
-													block,
-												],
+												detail: {
+													acfBlocksWithValidationErrors:
```
priethor, line 1181: See [my comment above/below](https://github.com/WordPress/secure-custom-fields/pull/281#discussion_r2651274916) about the name mismatch.

assets/src/js/pro/blocks-v3/components/block-edit.js
```diff
@@ -362,9 +416,13 @@ export const BlockEdit = ( props ) => {
 
 	// Listen for validation error events from other blocks
 	useEffect( () => {
-		const handleErrorEvent = () => {
-			lockPostSaving( clientId );
-			setShowValidationErrors( true );
+		const handleErrorEvent = ( event ) => {
+			// Only handle if this event is for this specific block
+			if ( clientId === event.detail.acfBlockWithValidationErrors ) {
```
priethor: The event passed by the hanlder seems to be called `acfBlocksWithValidationErrors`, in plural.
```suggestion
			if ( clientId === event.detail.acfBlocksWithValidationErrors ) {
```

includes/blocks.php
cbravobernal: The first one will create this HTML

`<link rel="stylesheet" id="acf-inline-editing-styles-css" href="http://localhost:8888/wp-content/plugins/secure-custom-fields/assets/build/css/pro/acf-styles-in-iframe-for-blocks.min.css?ver=1" media="all">`

The one you provided will create this one:

`<link rel="stylesheet" id="acf-inline-editing-styles-css" href="http://localhost:8888/wp-content/plugins/secure-custom-fields/assets/build/css/pro/acf-styles-in-iframe-for-blocks.min.css?ver=6.7.1" media="1">`

Let's use this code:
`wp_enqueue_style( 'acf-inline-editing-styles', $css_path, array(), ACF_VERSION, 'all' );`

and return: 

`<link rel="stylesheet" id="acf-inline-editing-styles-css" href="http://localhost:8888/wp-content/plugins/secure-custom-fields/assets/build/css/pro/acf-styles-in-iframe-for-blocks.min.css?ver=6.7.1" media="all">`

That way we cover the media property and the versioning as it should.

assets/src/js/pro/blocks-v3/components/block-edit.js
cbravobernal: Fixed. Thanks!

assets/src/js/_acf-validation.js
cbravobernal: Done!

priethor (approved):
General testing didn't give any visible errors; let's merge even if the solution is not ideal.

Thanks @cbravobernal! 🪨 ⭐ 

PS: we might need to skip codecov in PRs with the `Backport from upstream` label, at least until we have good general coverage.

cbravobernal:
> 
> PS: we might need to ignore codecov in PRs with the `Backport from upstream` label, at least until we have good general coverage.

Definitely 😅

PR #283, opened 2025-12-12
Fix checkbox toggle functionality

### LINKED ISSUE #282: Bug: Toggle All does not work and emits console error on tools page
The "Toggle All" checkbox does not work within the Tools > Export > "Select Field Groups" and "Select Post Types" fieldsets.

```
acf-input.min.js?ver=6a7ef76ae41206824f6c:1 Uncaught ReferenceError: $inputs is not defined
    at n.onClickToggle (acf-input.min.js?ver=6a7ef76ae41206824f6c:1:103505)
    at n.<anonymous> (acf.min.js?ver=767d85363a2695321d19:1:8868)
    at HTMLInputElement.i (jquery.min.js?ver=3.7.1:2:86755)
    at HTMLDivElement.dispatch (jquery.min.js?ver=3.7.1:2:40035)
    at v.handle (jquery.min.js?ver=3.7.1:2:38006)
```

SCF Version 6.7.0

jamieburchell:
Fixes https://github.com/WordPress/secure-custom-fields/issues/282

Also standardises the use of `var` and `$` instead of `const` and `jQuery` respectively and generally simplifies the logic.

### DIFF
--- a/assets/src/js/_acf-field-checkbox.js
+++ b/assets/src/js/_acf-field-checkbox.js
@@ -76,11 +76,9 @@
 		},
 
 		onClickToggle: function ( e, $el ) {
-			// Vars.
 			var $inputs = this.$inputs();
-			var hasUnchecked = $inputs.not( ':checked' ).length > 0;
-			$inputs.prop( 'checked', hasUnchecked ).trigger( 'change' );
-			$el.prop( 'checked', hasUnchecked );
+			var checked = $el.prop( 'checked' );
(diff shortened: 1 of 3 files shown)


priethor (approved):
Thanks for the contribution, @jamieburchell !

I can reproduce the issue in trunk and can confirm that this patch elegantly solves it.

I've added a couple of unit tests to prevent future regressions; they fail in trunk but pass in this PR, as expected.

PR #286, opened 2025-12-16
Fix PHP warnings when clone field encounters invalid field entries

priethor:
## What

Fix PHP warnings in the Clone field when the field array contains null or invalid entries.

Reported in the forum: https://wordpress.org/support/topic/php-warnings-207/

## Why
Users migrating from ACF to SCF, which includes PRO functionality, can encounter PHP warnings they didn't have before

## How
- Added defensive null/array check before accessing field properties in `acf_get_fields()` method
- Added unit tests to verify the fix and prevent regression

## Testing Instructions

- Create a field group with clone fields and verify no PHP warnings appear during normal usage
- Run `vendor/bin/phpunit --filter Test_ACF_Field_Clone`

### DIFF
--- a/includes/fields/class-acf-field-clone.php
+++ b/includes/fields/class-acf-field-clone.php
@@ -139,6 +139,12 @@ if ( ! class_exists( 'acf_field_clone' ) ) :
 			$count = count( $fields );
 			while ( $i < $count ) {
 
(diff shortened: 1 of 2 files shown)


cbravobernal (approved):
LGTM

PR #295, opened 2025-12-20
making other WC Order types (other than subscription) compatible with HPOS and custom fields

trajche:
Currently, other WC Order types than Subscription are not compatible (they don't show up in the edit page) with SCF and HPOS even though they show up in the field group rules when creating them. This PR fixes that.

### DIFF
--- a/includes/forms/WC_Order.php
+++ b/includes/forms/WC_Order.php
@@ -21,9 +21,10 @@ class WC_Order {
 	 * @since 6.5
 	 */
 	public function __construct() {
-		add_action( 'load-woocommerce_page_wc-orders', array( $this, 'initialize' ) );
-		add_action( 'load-woocommerce_page_wc-orders--shop_subscription', array( $this, 'initialize' ) );
-		add_action( 'woocommerce_update_order', array( $this, 'save_order' ), 10, 1 );
+        add_action( 'load-woocommerce_page_wc-orders', array( $this, 'initialize' ) );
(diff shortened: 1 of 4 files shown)


trajche:
In order to test you would have to create a custom WC order type and then **enable HPOS**. By default, it will allow you to select the custom WC order type from the SCF Field Groups but it will not show the meta boxes in the custom WC order view.

Here is sample code to register a WC order type called "Charges".
```
wc_register_order_type(
	'shop_order_charge',
	array(
		'label'                            => __( 'Charges', 'woocommerce' ),
		'capability_type'                  => 'shop_order',
		'public'                           => true,
		'hierarchical'                     => true,
		'exclude_from_orders_screen'       => false,
		'add_order_meta_boxes'             => true,
		'exclude_from_order_count'         => true,
		'exclude_from_order_views'         => false,
		'exclude_from_order_reports'       => false,
		'exclude_from_order_sales_reports' => false,
		'class_name'                       => 'WC_Order_Charge',
		'supports'                         => array( 'title' ),
		'rewrite'                          => true,
	)
);
```

And the Class would be something simple:
```
class WC_Order_Charge extends WC_Order {

	protected $object_type = 'order_charge';

	public function get_type() {
		return 'shop_order_charge';
	}
}
```

cbravobernal (approved):
Thanks for contributing!

LGTM, I will merge it and then update the @ since to 6.7.2.

Next time please use a branch! It will make editions much easier.

PR #302, opened 2025-12-26
JSON Schemas: Make them draft-07 compliant with fully working `$ref`s

priethor:
## What

Makes JSON schemas fully compliant with JSON Schema draft-07. This includes resolving the "Field fragments" into a single Field schema rather than composing it at runtime. In a nutshell, the main goal of draft-07 is leveraging `$ref` for reusability.

## Why

External validators and IDEs couldn't properly validate the schemas because of the `$ref` format and field property merging we were using for reusability and convenience: the field fragments that were being resolved at runtime are a practical way to reuse all the common properties all fields have, but that meant that the invididual schemas, while technically v07 compliant, were not capable of validating real, whole field JSON data if not using SCF's runtime builder to "make them full".

## How

On the one hand, reorganizing field schemas into a clearer fragment structure where:
  - `field-fragments/field-base.schema.json` contains shared definitions
  - `field-fragments/{category}/*.schema.json` contains type-specific properties.

Then, instead of merging the base + fragment at runtime as needed, we use the JSON Builder to generate a unified `field.schema.json` file, which is then used as any other plain schema. The generated file still cointains `$ref`s that are resolved at runtime because of the limited WP schema support: we _could_ resolve them _before_ generating the file, but this still maintains a good level of reusability in the full Field schema file (there will be 35 individual field definitions that must contain all properties, as we can't leverage `allOf` and other advanced schema options that we would also also need to resolve at runtime for the abilities API) while maintaining full draft-07 compliance.

To ensure the autogenerated `field.schema.json` file is kept in sync, this PR also adds a CI workflow that verifies the generated schema stays in sync with source fragments on every PR that touches schema files.

Finally, this also ensures external refs are properly resolved for WP's internal schema validation and replaces custom reference formats with standard JSON Schema v7 relative refs (e.g., `common.schema.json#/definitions/title`).

## Testing Instructions

1. Run `php bin/generate-field-schema.php --check`: it should report "in sync"
2. Modify any fragment file, run `--check` again: it should fail
3. Run `php bin/generate-field-schema.php` to regenerate
4. Validate `schemas/field.schema.json` with an external JSON Schema validator
5. Run `vendor/bin/phpunit tests/php/includes/SCFSchemaBuilderTest.php`: all tests should pass

### DIFF
--- a/.github/workflows/verify-field-schema.yml
+++ b/.github/workflows/verify-field-schema.yml
@@ -1,0 +1,38 @@
+name: Verify Field Schema
+
+on:
+  push:
+    branches:
+      - trunk
+      - 'branch/*'
+    paths:
+      - 'schemas/field-fragments/**'
+      - 'schemas/field.schema.json'
+      - 'bin/generate-field-schema.php'
+      - 'includes/class-scf-schema-builder.php'
+  pull_request:
+    branches:
+      - trunk
+      - 'branch/*'
+    paths:
+      - 'schemas/field-fragments/**'
+      - 'schemas/field.schema.json'
+      - 'bin/generate-field-schema.php'
+      - 'includes/class-scf-schema-builder.php'
+
+jobs:
+  verify-schema:
+    name: Verify field.schema.json is in sync
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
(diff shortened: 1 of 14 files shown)


priethor:
`field.schema.json` syncing validator works as expected 🙂 
image

PR #353, opened 2026-01-06
Fix dynamic update of field name

### LINKED ISSUE #350: Dynamically change the display of a field name when it is changed
Hello!

After the release of one of the latest versions (most likely 6.6.0), a small bug appeared.

On the field group editing page, each field has its own accordion. When collapsed, it has columns: label, name, and type.

Previously, when editing a field name, it would immediately change in the summary, but now this doesn't happen.

Image

YanMetelitsa:
## What
Fixes #350


cbravobernal (approved):
Nice catch

PR #355, opened 2026-01-14
Exclude bin/ directory from code coverage

kraftbj:
## Summary

- Excludes the `bin/` directory from Codecov coverage calculations

The `bin/` directory contains CLI and infrastructure scripts that make HTTP requests to external APIs (GitHub REST/GraphQL). These are integration scripts that are inherently difficult to unit test without extensive HTTP mocking.

See #354 as an example - the `codecov/patch` check passes (confirming the testable parsing logic has coverage), but `codecov/project` fails because the CLI orchestration code in `bin/` isn't covered.

## Test plan

- [ ] Verify Codecov no longer counts `bin/` files toward project coverage

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -39,7 +39,7 @@ ignore:
     - '*.min.css'
     - '*.min.js'
     - '.wp-env'
-    - 'bin/' # CLI/infrastructure scripts - integration code with external APIs
+    - 'bin' # CLI/infrastructure scripts - integration code with external APIs
     - 'build'
     - 'dist'
     - 'docs'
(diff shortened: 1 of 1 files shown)


ockham:
> See #354 as an example - the `codecov/patch` check passes (confirming the testable parsing logic has coverage), but `codecov/project` fails because the CLI orchestration code in `bin/` isn't covered.

Seems to be passing now though for that PR?

image

kraftbj:
Yeah, the last commit apparently changed the coverage so it wouldn't fail. Here's an example of the run where it did fail https://github.com/WordPress/secure-custom-fields/runs/60370170867

I'm not exactly sure why it differed so much between those two commits.

priethor (approved):

PR #356, opened 2026-01-14
Add milestone automation for merged PRs

### LINKED ISSUE #152: Release management: classify PRs in the right milestone to generate changelog
As of SCF 6.4, the milestone and changelog management is a manual process. We should consider adding a GitHub to add merged PRs to the next milestone, either upon PR merged (like the Gutenberg plugin does) or as part of the release process, to help with the changelog management.

kraftbj:
## Summary

Adds a GitHub workflow that assigns merged, unmilestoned pull requests targeting `trunk` to the explicitly configured next release milestone. The approach is inspired by [Gutenberg's project-management automation](https://github.com/WordPress/gutenberg/tree/trunk/packages/project-management-automation/lib/tasks/add-milestone), while keeping SCF's irregular and parallel release lines under maintainer control.

### Behavior

| Current state | Action |
| --- | --- |
| No milestone | Assign the open milestone named by the `NEXT_MILESTONE` repository variable |
| Open milestone | Keep it unchanged |
| Closed milestone | Keep it unchanged and add one idempotent warning comment |

The workflow reads the PR's live state, handles paginated GitHub API responses, runs only for merges into `trunk`, and uses only `issues: write` permission. It fails visibly rather than guessing when `NEXT_MILESTONE` is missing or invalid.

The release documentation now explains how to configure the target:

```sh
gh variable set NEXT_MILESTONE --body "X.Y.Z"
```

The repository is configured with open milestone `6.9.2` and `NEXT_MILESTONE=6.9.2`. Released milestones through `6.9.1` have been closed.

Closes #152

## Test plan

- [x] Exercise open, closed, duplicate-warning, assignment, missing-variable, and missing-target paths with mocked GitHub API responses
- [x] Parse the workflow as YAML
- [x] Run Prettier against the workflow
- [x] Run markdownlint against the release documentation
- [x] Run `git diff --check`
- [ ] Merge an unmilestoned PR into `trunk` and confirm it receives milestone `6.9.2`

## Use of AI Tools

Codex reviewed the original workflow, implemented the requested update under maintainer direction, and exercised its behavior with mocked GitHub API responses. The resulting changes were reviewed and validated locally before push.


ockham:
I'm not entirely sure about the closed milestone behavior. I'm not sure I like the idea that an aspect of a PR that I set manually -- the milestone -- might be automatically changed when I merge it. I guess it's meant to help me in case I forget to assign a new milestone if the previous one was already closed, but to me, it seems like it actually removes confidence, because now it feels like I'll always have to double-check.

Maybe I'm biased by my experience with Gutenberg, where the milestone is closed when the RC is published, meaning that any bugfixes that need to go in afterwards and before the stable version is released will need to be assigned a closed milestone. I guess the release workflow is different for SCF (?)

Anyway, how about we don't update the milestone if it's closed and instead add a comment to notify the PR author to make sure they're aware?

kraftbj:
I'm fine backing off the closed milestone process. It may be a solution in search of a problem.

PR #368, opened 2026-03-01
Fix TypeError: Cannot read properties of undefined (reading 'localeCompare')

EarthmanWeb:
## Summary

Fixes an `Uncaught TypeError: Cannot read properties of undefined (reading 'localeCompare')` thrown on every admin page load from `scf-admin.min.js`.

## Problem

Both `admin-commands.js` and `custom-post-type-commands.js` read `window.acf.data.wp_version` and immediately call `.localeCompare()` on it to determine if the WordPress version is 6.9+. However, `wp_version` is only added to `acf.data` conditionally in `includes/assets.php` (line 622), inside a block that checks for Gutenberg/block editor context. On classic editor pages and other admin screens where that condition isn't met, `wp_version` is `undefined`, causing the TypeError.

**Error:**
```
Uncaught TypeError: Cannot read properties of undefined (reading 'localeCompare')
    at h (scf-admin.min.js?ver=6.8.0:1:6456)
```

This error fires on every admin page load via `requestIdleCallback`.

## Fix

Added a fallback `|| '0'` when reading `window.acf.data.wp_version` in both files:

```js
// Before (crashes when wp_version is undefined)
const wpVersion = window.acf.data.wp_version;

// After (safely falls back to '0', treating unknown versions as pre-6.9)
const wpVersion = window.acf.data.wp_version || '0';
```

When `wp_version` is undefined, the fallback `'0'` ensures `localeCompare('6.9')` returns `< 0`, so all commands are registered (pre-6.9 behavior). This is the safe default — registering extra commands is harmless, while crashing is not.

## Files Changed

- `assets/src/js/commands/admin-commands.js` — Added `|| '0'` fallback (line 216)
- `assets/src/js/commands/custom-post-type-commands.js` — Added `|| '0'` fallback (line 41)
- `tests/js/commands/admin-commands.test.js` — Added test case for `undefined` wp_version
- `tests/js/commands/custom-post-type-commands.test.js` — Added test case for `undefined` wp_version

## Test Plan

- [x] Verified fix resolves the error on a production WordPress 6.9.1 site
- [x] Confirmed all existing commands still register correctly
- [ ] `npm run test:unit` passes with new undefined version test cases
- [ ] Verify on WP 6.9+ that view commands are correctly skipped when `wp_version` is available

### DIFF
--- a/assets/src/js/commands/admin-commands.js
+++ b/assets/src/js/commands/admin-commands.js
@@ -213,7 +213,7 @@ const registerAdminCommands = () => {
 	};
 
 	// WordPress 6.9+ adds Command Palette commands for all admin menu items.
-	const wpVersion = window.acf.data.wp_version;
+	const wpVersion = window.acf.data.wp_version || '0';
 	const isWp69Plus =
 		wpVersion.localeCompare( '6.9', undefined, { numeric: true } ) >= 0;
 
--- a/assets/src/js/commands/admin-commands.js
+++ b/assets/src/js/commands/admin-commands.js
@@ -31,7 +31,7 @@ import {
  * Register admin commands for SCF
  */
 const registerAdminCommands = () => {
-	if ( ! dispatch( 'core/commands' ) || ! window.acf?.data ) {
+	if ( ! dispatch( 'core/commands' ) || ! window.acf?.data?.wp_version ) {
 		return;
 	}
 
@@ -213,7 +213,7 @@ const registerAdminCommands = () => {
 	};
 
 	// WordPress 6.9+ adds Command Palette commands for all admin menu items.
-	const wpVersion = window.acf.data.wp_version || '0';
+	const wpVersion = window.acf.data.wp_version;
 	const isWp69Plus =
 		wpVersion.localeCompare( '6.9', undefined, { numeric: true } ) >= 0;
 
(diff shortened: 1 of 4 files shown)


ockham:
Good spot, thank you!

As for the fix, I think it'd be better if we bailed early if `window.acf.data.wp_version` isn't available (rather than defaulting the WP version number to `0`).

Turns out that we already have checks that look like they're supposed to cover cases like these: https://github.com/WordPress/secure-custom-fields/blob/8b53724ba29c14ab80e42701e89653d69e7c384e/assets/src/js/commands/admin-commands.js#L34-L36

and 

https://github.com/WordPress/secure-custom-fields/blob/8b53724ba29c14ab80e42701e89653d69e7c384e/assets/src/js/commands/custom-post-type-commands.js#L28-L34

They just don't seem to quite cover the case where `wp_version` isn't present 🤔 
Maybe you can tweak those conditionals?

ockham:
Looking a bit more into this had me discover a related/underlying issue: When the command palette is invoked (Cmd+K) from a non-editor wp-admin screen, SCF options aren't added there. (This is due to the [conditionals I mentioned](https://github.com/WordPress/secure-custom-fields/pull/368#issuecomment-4024705149).)

| Editor | Other wp-admin |
|-|-|
| image | image |

ockham:
Closing in favor of https://github.com/WordPress/secure-custom-fields/pull/397.

PR #372, opened 2026-03-09
Add GHA annotations for `phpcs-changed` output

ockham:


ockham:
Looks like it's working 🎉 With https://github.com/WordPress/secure-custom-fields/pull/372/commits/5ba49e978c48ca2c474bfc619b0ccb1fbee45746 applied:

image

PR #379, opened 2026-03-10
Prepare 6.8.1 Release

ockham:
*Release Date 11th March 2026*

* Backports from 6.7.1*

- Security - User field AJAX queries now enforce field-configured role restrictions and validate search permissions.
- Security - Post Object, Relationship, and Page Link field AJAX queries now enforce field-configured restrictions for post status, post type, and taxonomy.
- Site Health - Track blocks using auto inline editing.

### DIFF
--- a/docs/code-reference/META.md
+++ b/docs/code-reference/META.md
@@ -10,6 +10,12 @@ This file tracks code elements that need documentation.
 - `acf/bindings/field_not_supported_message`
 - `acf/blocks/binding_value`
 
+## abilities/class-scf-field-abilities.php
+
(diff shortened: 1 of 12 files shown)


ockham:
Okay so PR generation isn't working for point releases -- which makes sense: The `release/6.8` branch was originally created for the 6.8.0 release and then merged into `trunk`, so the commits from back then are already present there.

We'll need to reconsider our workflows. It does make sense to have a `release/x.y` branch for each `x.y` series, but it can't be reused for PRs for `x.y.z` point releases.

PR #382, opened 2026-03-10
Harden radio save_other_choice lookup for JSON fields

### LINKED ISSUE #381: Radio save_other_choice triggers warnings for JSON-only or unidentified fields
## Summary
When a Radio field has `save_other_choice` enabled, `update_value()` can emit PHP warnings if the field is JSON-only (no DB `ID`) or missing identifiers.

## Affected area
- `includes/fields/class-acf-field-radio.php`
- Method: `acf_field_radio::update_value()`

## Problem
Current logic assumes the field always has an `ID`/`key` and that `acf_get_field()` returns a valid array with `ID`. In JSON/local-field scenarios, that may not be true, which can produce warnings (undefined array key / invalid array offset access).

## Steps to reproduce
1. Create a Radio field with `save_other_choice = 1`.
2. Use a JSON/local field config (or a field array without persisted DB `ID`).
3. Save a custom value not present in `choices`.
4. Observe PHP warnings during `update_value()`.

## Expected behavior
No warnings should be emitted. Value should still be returned/saved, and save-other-choice should only run when a valid persisted field is resolvable.

## Proposed fix
- Build selector defensively:
  - use `ID` if available
  - else use `key` if available
  - bail if both are missing
- Bail if `acf_get_field()` returns non-array or no `ID`
- Ensure `choices` is an array before appending custom values

## Tests
Add regression tests in:
`tests/php/includes/fields/test-class-acf-field-radio.php`
- `test_update_value_save_other_choice_handles_json_field_without_id`
- `test_update_value_save_other_choice_handles_missing_field_identifier`

## Patch reference
Ready in commit: `94d21e5`

meravi:
## Summary
This hardens `acf_field_radio::update_value()` when `save_other_choice` is enabled and the runtime field is JSON-only or missing identifiers.

## Changes
- Guard selector resolution (`ID` -> `key` -> bail early).
- Bail if `acf_get_field()` fails or returns a field without DB `ID`.
- Normalize `choices` to an array before appending custom values.
- Add regression tests for JSON-only/missing-identifier edge cases.

## Tests
- Added:
  - `test_update_value_save_other_choice_handles_json_field_without_id`
  - `test_update_value_save_other_choice_handles_missing_field_identifier`

Fixes #381

### DIFF
--- a/includes/fields/class-acf-field-radio.php
+++ b/includes/fields/class-acf-field-radio.php
@@ -321,8 +321,6 @@ if ( ! class_exists( 'acf_field_radio' ) ) :
 		 * @type    filter
 		 * @since   ACF 3.6
 		 * @date    23/01/13
-		 * @todo    Fix bug where $field was found via json and has no ID
-		 *
 		 * @param   $value - the value which will be saved in the database
 		 * @param   $post_id - the post_id of which the value will be saved
 		 * @param   $field - the field array holding all the field options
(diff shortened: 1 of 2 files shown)


cbravobernal (commented):
Verified locally:

- The two new tests reproduce the reported warning against current trunk (`Undefined array key "ID"` in `class-acf-field-radio.php:347`) and pass with this fix.
- Full PHPUnit suite (2230 tests), PHPStan, and phpcs all green on this branch.

The defensive selector build (ID → key → bail) and the `acf_get_field()` return check both look right, and behavior for valid persisted fields is unchanged.

*Verification run via Claude Code, reviewed by @cbravobernal.*

tapheret2 (commented):
### Review: Harden radio save_other_choice lookup for JSON fields

**Files:** `includes/fields/class-acf-field-radio.php`, `tests/php/includes/fields/test-class-acf-field-radio.php`  
**Size:** +50 / -9

**Notes**

**Nits**

Thanks @meravi — independent review pass on the diff.

PR #383, opened 2026-03-10
Select/Checkbox save_custom/save_options triggers warnings for JSON-only or unidentified fields

meravi:
## Summary
When custom option saving is enabled, Select (`save_options`) and Checkbox (`save_custom`) can emit PHP warnings if the runtime field is JSON-only (no DB `ID`) or missing identifiers.

## Affected areas
- `includes/fields/class-acf-field-select.php` (`acf_field_select::update_value`)
- `includes/fields/class-acf-field-checkbox.php` (`acf_field_checkbox::update_value`)

## Problem
Current logic assumes a valid field identifier (`ID` or `key`) and a successful `acf_get_field()` lookup with a persisted `ID`. In JSON/local-field scenarios this may be false, which can lead to warnings (undefined array key / invalid array access) while trying to append custom choices.

## Steps to reproduce
1. Configure a Select field with `save_options = 1` (or Checkbox with `save_custom = 1`).
2. Use field data coming from JSON/local config (or with missing persisted `ID`).
3. Save a custom value not present in `choices`.
4. Observe warnings during `update_value()`.

## Expected behavior
No warnings should be emitted. Value should still be returned/saved, and custom-choice persistence should only run when a valid persisted field can be resolved.

## Proposed fix
- Resolve selector defensively (`ID` -> `key` -> bail).
- Bail when lookup cannot run or `acf_get_field()` result is invalid/no `ID`.
- Normalize `choices` to an array before appending custom values.

## Tests
Added regression tests:
- `tests/php/includes/fields/test-class-acf-field-select.php`
  - `test_update_value_save_options_handles_json_field_without_id`
  - `test_update_value_save_options_handles_missing_field_identifier`
- `tests/php/includes/fields/test-class-acf-field-checkbox.php`
  - `test_update_value_save_custom_handles_json_field_without_id`
  - `test_update_value_save_custom_handles_missing_field_identifier`

## Patch reference
Commit: `34cfcb7`
Branch: `fix/select-checkbox-save-options-guards`

### DIFF
--- a/includes/fields/class-acf-field-checkbox.php
+++ b/includes/fields/class-acf-field-checkbox.php
@@ -499,20 +499,32 @@ if ( ! class_exists( 'acf_field_checkbox' ) ) :
 			// select -> update_value()
 			$value = acf_get_field_type( 'select' )->update_value( $value, $post_id, $field );
 
-			// save_other_choice
-			if ( $field['save_custom'] ) {
+			// save_custom
+			if ( ! empty( $field['save_custom'] ) && is_array( $value ) ) {
 
 				// get raw $field (may have been changed via repeater field)
 				// if field is local, it won't have an ID
-				$selector = $field['ID'] ? $field['ID'] : $field['key'];
-				$field    = acf_get_field( $selector );
-				if ( ! $field ) {
-					return false;
+				$selector = '';
+				if ( ! empty( $field['ID'] ) ) {
+					$selector = $field['ID'];
+				} elseif ( ! empty( $field['key'] ) ) {
+					$selector = $field['key'];
 				}
 
-				// bail early if no ID (JSON only)
-				if ( ! $field['ID'] ) {
+				// bail early if we have no selector for lookup
+				if ( ! $selector ) {
 					return $value;
+				}
+
+				$field = acf_get_field( $selector );
+
+				// bail early if lookup failed or field is JSON only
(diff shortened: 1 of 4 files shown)


cbravobernal (commented):
Verified locally:

- The four new tests reproduce the reported warnings against current trunk (e.g. `Undefined array key "ID"` in `class-acf-field-select.php:585`) and pass with this fix.
- Full PHPUnit suite (2232 tests), PHPStan, and phpcs all green on this branch.

One behavior change worth calling out explicitly for the record: in the checkbox field, the failed-lookup path previously did `return false;` (dropping the submitted value entirely), and now returns `$value`. That's an improvement — consistent with the select/radio handling — but it is a behavior change, not just a warning fix.

Also +1 on adding the `is_array( $value )` guard before the save_custom loop.

*Verification run via Claude Code, reviewed by @cbravobernal.*

tapheret2 (commented):
### Review: Select/Checkbox save_custom/save_options triggers warnings for JSON-only or unidentified fields

**Files:** `includes/fields/class-acf-field-checkbox.php`, `includes/fields/class-acf-field-select.php`, `tests/php/includes/fields/test-class-acf-field-checkbox.php`, `tests/php/includes/fields/test-class-acf-field-select.php`  
**Size:** +98 / -11

**Notes**

**Nits**

Thanks @meravi — independent review pass on the diff.

PR #385, opened 2026-03-11
Add release documentation

ockham:
Based on my notes from a video call with @cbravobernal, and recent hands-on experience 😅

### DIFF
--- a/docs/contributing/releases.md
+++ b/docs/contributing/releases.md
@@ -1,6 +1,6 @@
 # Releasing a New Version
 
-Read the following guide to learn how to created and publish a new release of Secure Custom Fields.
+Read the following guide to learn how to create and publish a new release of Secure Custom Fields. (This is only relevant for project maintainers.)
 
 ## Prerequisites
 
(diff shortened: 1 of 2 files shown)


docs/contributing/releases.md
```diff
@@ -0,0 +1,101 @@
...
+svn ci -m “Prepare release x.y.z”
+```
+
+### Set rollout strategy
+
+After committing, you will receive an email asking you to select the rollout strategy for the new version of the plugin. Follow the instructions in that email. You can choose between "Immediate (default)" and "Manual updates only (24 hours)". You might want to select the latter option, as it gives you some more time to ask others to manually download and test the plugin from https://wordpress.org/plugins/secure-custom-fields/.
+
+### Set stable tag
+
+Finally, don't forget to update the stable tag in `readme.txt`. You need to do so both in the git repository, and in SVN. In both cases, it's sufficient to commit the change to `trunk` only. For completeness' sake, you should also create an `x.y.z` tag in git (and have it point to the commit that bumped the stable tag).
```
ockham: Needs more detail, especially around GH tag creation. (Create via GH Releases UI?)

Also, branch handling isn't quite clear. `release/x.y.z` branch should be kept around so that stable tag commit can sit on top of it.

PR #392, opened 2026-03-21
Fix _load_textdomain_just_in_time notice on WordPress 6.7+

### LINKED ISSUE #252: Function _load_textdomain_just_in_time was called incorrectly (WP version 6.7 and above)
the following notice for secure custom fields is now displayed for wordpress version 6.7 and above:

Notice: Function _load_textdomain_just_in_time was called incorrectly. Loading translation for the secure-custom-fields domain was activated too early. This is usually an indication that some code in the plugin or the theme encounters something too early. Translations should be loaded during the init action or later. Read Debugging in WordPress for more information. (This message was added in version 6.7.0.)

trajche:
## Summary
- Replaces `load_textdomain()` with `load_plugin_textdomain()` in `acf_load_textdomain()` for proper integration with the just-in-time translation loading system introduced in WordPress 6.7.
- This fixes the `_load_textdomain_just_in_time was called incorrectly` notice that appears on WordPress 6.7+ installations.

Fixes #252

## Test plan
- [ ] Install SCF on WordPress 6.7+ with `WP_DEBUG` and `WP_DEBUG_DISPLAY` enabled
- [ ] Verify no `_load_textdomain_just_in_time` notice appears for the `secure-custom-fields` domain
- [ ] Verify translations still load correctly on localized WordPress installations

### DIFF
--- a/includes/l10n.php
+++ b/includes/l10n.php
@@ -103,20 +103,10 @@ function acf_get_locale() {
  */
 function acf_load_textdomain( $domain = 'secure-custom-fields' ) {
 
-	/**
-	 * Filters a plugin's locale.
-	 *
-	 * @date    8/1/19
-	 * @since   ACF 5.7.10
-	 *
-	 * @param   string $locale The plugin's current locale.
-	 * @param   string $domain Text domain. Unique identifier for retrieving translated strings.
-	 */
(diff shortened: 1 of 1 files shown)


trajche:
This was not an issue with SCF. It was caused by other plugins registering options page/acf fields too early, which initalized SCF. By wrapping it in `acf/init` action, it works just fine.

PR #397, opened 2026-03-26
Fix command palette in wp-admin

ockham:
Fixes the `Uncaught TypeError: Cannot read properties of undefined (reading 'localeCompare')` error thrown on `wp-admin`, and makes sure that commands are registered correctly.

[As of WP 6.9, the command palette is available from wp-admin](https://make.wordpress.org/core/2025/11/25/wordpress-6-9-field-guide/#site-editorblock-editor), so it makes sense for SCF to register its commands there, too.

Additionally, this PR:
- removes the superfluous `description` fields from commands, as those [aren't supported by the Commands API](https://github.com/WordPress/gutenberg/blob/924fe383e22b25748e1c9292089bba007ef3c475/packages/commands/README.md#types-of-commands).
- removes the `context` field, as it was set to `admin`, which is not a supported value for that field. For a full list, see [here](https://github.com/WordPress/gutenberg/tree/74725e041ccad918f76160d68a82bfb00be27f64/packages/commands#contextual-commands). (None of the possible values seem to make sense for SCF related commands.)

Closes #368.

## TODO

- [x] Check if `custom-post-type-commands.js` needs modifications
- [x] Make sure CPT related commands are registered from wp-admin screens
  - [x] Need to figure out a way to expose list of ACF-registered CPTs so it's visible from wp-admin
- [x] Remove WP version check and register CPT related commands only if an equivalent command hasn't been registered yet (see `admin-commands.js`).
- [x] Make sure only commands available to the current user (based on permissions) are registered, see https://github.com/WordPress/secure-custom-fields/blob/2875e2c823e7058cb0b2c47a26c9bb6aaca3425e/includes/admin/admin-commands.php#L20-L23
  - [x] *Bug* (present in `trunk`): "Edit post type" command is available to non-admin users. _Filed #402 for this._
  - [x] Client-side check? (No need to add new restrictions to the /types endpoint on the server side.)
- [x] Remove now-obsolete `customPostTypes` var from `includes/admin/admin-commands.php`.
  - [x] Only enqueue `scf-commands-custom-post-types` script if there are AFC-created post types?
- [ ] Add e2e coverage for CPT related commands:
  - [x] Called from post editor, called from wp-admin
  - [ ] Called as admin, called as user with lesser permissions
- [x] Audit `context` field (present on some commands)
- [x] Figure out why e2e tests fail in 6.8 (see CI)
- [x] Add e2e coverage for command palette called from editor
- [x] Remove temporary changes to e2e test matrix in CI

## Use of AI Tools

Claude Code helped add the `scf_post_id` field to the `types` endpoint, and created the original E2E test.


tests/e2e/fixtures.js
```diff
@@ -177,13 +177,13 @@ const test = wpTest.extend( {
...
-			const branchClass = [ ...document.body.classList ].find( ( c ) =>
-				c.startsWith( 'branch-' )
+			const versionClass = [ ...document.body.classList ].find( ( c ) =>
+				c.startsWith( 'version-' )
 			);
-			if ( ! branchClass ) {
+			if ( ! versionClass ) {
 				return true;
 			}
-			const match = branchClass.match( /branch-(\d+)-(\d+)/ );
```
ockham, line 186: For WP 7.0, `branchClass` is `branch-7`, meaning that there won't be a match. Fortunately, there's the more "complete" `version-7-0` class.

includes/rest-api/class-acf-rest-types-endpoint.php
```diff
@@ -183,6 +183,20 @@ public function register_extra_fields() {
...
+			array(
+				'get_callback' => array( $this, 'get_scf_post_id' ),
+				'schema'       => array(
+					'description' => __( 'The SCF internal post ID that defines this post type, or null if not managed by SCF.', 'secure-custom-fields' ),
+					'type'        => array( 'integer', 'null' ),
+					'context'     => array( 'view', 'edit' ),
+					'readonly'    => true,
+				),
+			)
+		);
```
ockham, line 199: TODO: Verify that ACF's internal post type IDs aren't considered sensitive data (and/or add permissions guard).

includes/rest-api/class-acf-rest-types-endpoint.php
```diff
@@ -225,6 +239,27 @@ public function get_scf_fields( $post_type_object ) {
 		return $field_groups_data;
 	}
 
+	/**
+	 * Get the SCF internal post ID for a post type.
+	 *
+	 * @since SCF 6.8.3
```
ockham, line 245: Maybe this change is big enough to warrant a 6.9.0.

assets/src/js/commands/custom-post-type-commands.js
```diff
@@ -5,130 +5,105 @@
...
+				postType.slug,
 				postType.name,
-				postType.label,
 			],
 			callback: ( { close } ) => {
-				document.location = addQueryArgs( adminUrl + 'post.php', {
+				document.location = addQueryArgs( 'post.php', {
 					post: postType.id,
 					action: 'edit',
 				} );
```
ockham, line 131: The "Edit Post Type" command might belong into `admin-commands.js`. (This isn't editing e.g. a Movie post, but for editing properties of the Movie post _type_. It lives under SCF's "Post Types" menu, which itself has a command in `admin-commands.js`.)

assets/src/js/commands/custom-post-type-commands.js
```diff
@@ -5,131 +5,119 @@
...
 		return;
 	}
 
-	const commandStore = dispatch( 'core/commands' );
-	const adminUrl = window.acf.data.admin_url || '';
-	const postTypes = window.acf.data.customPostTypes;
+	const postTypes = await resolveSelect( 'core' ).getPostTypes( {
+		per_page: -1,
+		source: 'scf',
+	} );
```
ockham, line 30: Do we need to filter for "active" SCF CPTs?

assets/src/js/commands/custom-post-type-commands.js
ockham: Filed https://github.com/WordPress/secure-custom-fields/issues/402 for this.

assets/src/js/commands/custom-post-type-commands.js
ockham: https://github.com/WordPress/secure-custom-fields/pull/397/commits/6fd3b892ea3460f6227d78f86f8b2fa2ffd2b0db

PR #399, opened 2026-03-26
Fix media modal crash on WP 7.0 due to attachment filters wrapper change

### LINKED ISSUE #396: E2E tests fail when run against WP `trunk`
End-to-end tests are currently failing when run against WP `trunk`. This seems to be happening when clicking on the "Add Image" button for an image field:

Image

The console shows the following error:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'all')
    at n.customizeFilters
```

The relevant line is https://github.com/WordPress/secure-custom-fields/blob/b448e17a01af46f78a9b5186ef6256ece1e83c53/assets/src/js/_acf-media.js#L378

kraftbj:
WP core commit [61757](https://core.trac.wordpress.org/changeset/61757) wrapped the media attachment filters `<select>` and `<label>` in a container `<div>` for accessibility. This changed `toolbar.get('filters')` from returning the `AttachmentFilters` view directly to returning a wrapper `wp.media.View`, causing `customizeFilters()` to crash with:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'all')
    at n.customizeFilters
```

This broke the media modal for any SCF image/file field on WP 7.0.

The fix resolves the actual `AttachmentFilters` view by checking whether the returned view has a `.filters` property. If not (WP 7.0+), it traverses the wrapper's subviews to find the `AttachmentFilters` instance. This is compatible with both WP < 7.0 and WP 7.0+.

Tested on both WP 6.9.4 and WP 7.0-RC2 with the full E2E suite passing.

Fixes #396
See https://core.trac.wordpress.org/ticket/64948

## Use of AI Tools

This PR was authored with assistance from Claude Code (Anthropic). The fix was developed after debugging the root cause using Playwright-based investigation of the media modal behavior on WP 7.0-RC2.


ockham:
I was going to wait a bit longer for https://core.trac.wordpress.org/ticket/64948 to be potentially fixed before WP 7.0 is released 🤔

kraftbj:
That's fine! I figure it's quick to merge and ship a quick compat release if we need to this way.

cbravobernal (commented):
The approach looks right and is adaptive to both structures: on WP < 7.0 (and on a future core where `toolbar.get( 'filters' )` returns the `AttachmentFilters` view directly again), the `! filtersView.filters` check falls through and nothing changes; on WP 7.0 the wrapper's subviews are searched.

One edge worth hardening while we're here: if `toolbar.get( 'filters' )` returns `undefined`, or no subview exposes `.filters`, `filters` remains the wrapper (or `undefined`) and the function still crashes a few lines later at `filters.filters.all`. A small guard after the lookup would make this bulletproof:

```js
if ( ! filters || ! filters.filters ) {
	return;
}
```

With that, the media modal degrades gracefully (default filters, no SCF customization) instead of breaking if core moves the view again.

*Review run via Claude Code, reviewed by @cbravobernal.*

cbravobernal (approved):
LGTM!

PR #401, opened 2026-04-02
E2E Tests: Fix post types setup plugin

ockham:
While working on https://github.com/WordPress/secure-custom-fields/pull/397, I needed to change the `add_new` label on the custom post type (CPT) used for tests. However, this change didn't seem to take effect on the frontend.

This was due to the way the test CPT was registered: It was persisted to the DB, with another DB option set as a flag so the test wouldn't attempt to recreate the CPT and instead use the persisted one. As a result, any changes made programmatically to the CPT didn't take effect.

This PR changes the test setup so the CPT is created on the fly by the plugin (and not persisted to the DB). 

## Use of AI Tools

Claude Code was used to investigate the issue, and to write the fix.
(This is only fair: The issue was originally also produced by Claude Code.)

### DIFF
--- a/tests/e2e/plugins/scf-test-setup-post-types.php
+++ b/tests/e2e/plugins/scf-test-setup-post-types.php
@@ -6,15 +6,6 @@
  * Author: SCF Testing
  *
  * @package wordpress/secure-custom-fields
- *
- * IMPORTANT NOTE:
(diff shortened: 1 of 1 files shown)


ockham:
Switching back to draft mode. I had another idea I wanted to try.

ockham:
Closing in favor of #403.

PR #404, opened 2026-04-08
Skip Command Palette e2e tests on older WP versions

ockham:
E2E tests are currently run against WP 6.2 upon commit to `trunk`. Command Palette tests are failing, as the Command Palette was only introduced in WP 6.3. We thus need to skip those tests on older WP versions.

Additionally, this PR makes the test more resilient against minor copy changes across versions that would previously cause tests to fail.

## Use of AI Tools

Claude Code was used to select the command palette input field in a label-agnostic way that would work across different WP versions.

### DIFF
--- a/tests/e2e/command-palette.spec.ts
+++ b/tests/e2e/command-palette.spec.ts
@@ -12,6 +12,15 @@ test.describe( 'Command Palette', () => {
 		await requestUtils.deactivatePlugin( 'secure-custom-fields' );
 	} );
 
+	test.beforeEach( async ( { admin, page } ) => {
+		await admin.visitAdminPage( 'index.php' );
+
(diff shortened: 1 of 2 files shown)


tests/e2e/command-palette.spec.ts
```diff
@@ -32,9 +41,10 @@ test.describe( 'Command Palette', () => {
 					// Open the command palette via keyboard shortcut.
 					await page.keyboard.press( 'ControlOrMeta+k' );
 
-					const input = page.getByRole( 'combobox', {
-						name: 'Search commands and settings',
```
ockham, line 36: This used to be `Type a command or search` in WP 6.3, so a label-agnostic way of locating the input field was needed.

PR #406, opened 2026-04-10
Fix block editor rendering: query decoding and StrictMode

adrianmomorales:
## Summary

Two fixes for ACF block rendering in the Gutenberg editor:

### 1. Decode `$query` when sent as JSON string (`acf_ajax_fetch_block`)

The block editor JS may serialize the `query` parameter as a JSON string, but `acf_ajax_fetch_block()` assumes it is already a PHP array. On **PHP 8.4** this causes a fatal TypeError at line 1099:

```
Cannot access offset of type string on string
```

The `$block` and `$context` parameters are already decoded with `json_decode()` (lines 1040-1046), but `$query` is not. This adds the same treatment.

### 2. Enable `StrictMode` flag in localized block editor data (gated on WP 6.9+)

The block editor JS already has code to handle React 18 StrictMode (mount → unmount → remount), gated behind `acf.get('StrictMode')`:

```js
// acf-pro-blocks.js — V.setState()
(this.subscribed || acf.get('StrictMode')) && super.setState(e);
```

But the flag is **never set** — `acf.get('StrictMode')` always returns `null`. Without it, when React StrictMode remounts a block component, `this.subscribed` is `false` (set during unmount) and `super.setState()` is skipped. The block form HTML is fetched successfully but never rendered — the block shows an infinite spinner.

This adds `'StrictMode' => version_compare( $GLOBALS['wp_version'], '6.9', '>=' )` to the localized data so the flag is only active where React StrictMode is actually used.

### PHPUnit tests

Added tests covering the query decoding logic: valid JSON string, invalid JSON fallback, array passthrough, non-array JSON values, and empty strings.

## Steps to reproduce

1. Use WordPress 6.9+ (React 18 StrictMode in the block editor)
2. Register any ACF block with `acf_block_version: 2` and `mode: edit`
3. Insert the block in the editor
4. Block shows infinite spinner (fix 2), or on PHP 8.4 a 500 error (fix 1)

## Test plan

- [ ] Insert an ACF block in the editor on PHP 8.4 — should render form without errors
- [ ] Insert an ACF block in the editor on WordPress 6.9+ — should render form without infinite spinner
- [ ] Verify block preview mode works correctly
- [ ] Verify block frontend rendering is unaffected

> **Note:** This replaces #391 which was closed when the fork was accidentally deleted. All review feedback from that PR has been incorporated.

### DIFF
--- a/includes/blocks.php
+++ b/includes/blocks.php
@@ -914,7 +914,7 @@ function acf_enqueue_block_assets() {
 		array(
 			'blockTypes' => array_values( $block_types ),
 			'postType'   => get_post_type(),
-			'StrictMode' => true,
+			'StrictMode' => version_compare( $GLOBALS['wp_version'], '6.9', '>=' ),
 		)
 	);
 
--- a/includes/blocks.php
+++ b/includes/blocks.php
@@ -899,6 +899,10 @@ function acf_enqueue_block_assets() {
 		)
 	);
 
+	// React StrictMode is only active in the block editor on WP 6.9+.
+	global $wp_version;
+	$is_wp_69_plus = version_compare( $wp_version, '6.9', '>=' );
+
 	// Get block types.
 	$block_types = array_map(
 		function ( $block ) {
@@ -914,7 +918,9 @@ function acf_enqueue_block_assets() {
 		array(
 			'blockTypes' => array_values( $block_types ),
 			'postType'   => get_post_type(),
-			'StrictMode' => version_compare( $GLOBALS['wp_version'], '6.9', '>=' ),
+			// Key is intentionally PascalCase to match `acf.get('StrictMode')` in the
+			// compiled `acf-pro-blocks.js`. Do not normalize to camelCase.
+			'StrictMode' => $is_wp_69_plus,
 		)
 	);
 
@@ -987,7 +993,37 @@ function acf_enqueue_block_type_assets( $block_type ) {
 }
 
 /**
+ * Decodes the block `query` ajax arg, which may arrive as either an array
+ * or a JSON-encoded string depending on how the block editor JS serializes
+ * the request. Falls back to an empty array when decoding fails or yields
+ * a non-array value.
+ *
+ * @since SCF 6.5.4
+ *
+ * @param mixed $query The raw query arg.
+ * @return array
+ */
+function acf_decode_block_query_arg( $query ) {
+	if ( is_array( $query ) ) {
+		return $query;
+	}
+
+	if ( ! is_string( $query ) ) {
+		return array();
+	}
+
+	$decoded = json_decode( wp_unslash( $query ), true );
+
+	return is_array( $decoded ) ? $decoded : array();
+}
+
+/**
  * Handles the ajax request for block data.
+ *
+ * The `query` arg may arrive as a JSON-encoded string instead of an array
+ * (the block editor JS may serialize it that way). It is normalized via
+ * `acf_decode_block_query_arg()` to avoid a fatal TypeError on PHP 8.4
+ * when accessing offsets on a string.
  *
  * @since   ACF 5.7.13
  *
@@ -1032,15 +1068,11 @@ function acf_ajax_fetch_block() {
 	$raw_context = $args['context'];
 	$post_id     = $args['post_id'];
 
-	// Decode query if sent as a JSON string instead of an array.
-	// The block editor JS may serialize the query parameter as JSON,
-	// which causes a fatal TypeError on PHP 8.4 when accessing offsets.
-	if ( is_string( $query ) ) {
-		$query = json_decode( wp_unslash( $query ), true );
-		if ( ! is_array( $query ) ) {
-			$query = array();
-		}
-	}
+	// `$query` flows through `acf_sanitize_request_args()` (so a JSON string
+	// is `wp_kses`-filtered before we see it). Safe today because callers only
+	// boolean-test keys like `preview`/`form`/`validate`; revisit if a string
+	// value ever gets read out of `$query`.
+	$query = acf_decode_block_query_arg( $query );
 
 	// Bail early if no block.
 	if ( ! $block ) {
--- a/includes/blocks.php
+++ b/includes/blocks.php
@@ -1028,6 +1028,9 @@ function acf_ajax_fetch_block() {
 
 	$block       = $args['block'];
 	$query       = $args['query'];
+	$client_id   = $args['clientId'];
+	$raw_context = $args['context'];
+	$post_id     = $args['post_id'];
 
 	// Decode query if sent as a JSON string instead of an array.
 	// The block editor JS may serialize the query parameter as JSON,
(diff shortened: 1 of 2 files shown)


adrianmomorales:
Hi! I've already linked my GitHub and WordPress.org accounts. Could a maintainer please add the `props-bot` label to refresh the contributor list? Thanks!

cbravobernal (commented):
## Summary

Approve direction — fixes are small, targeted, correct. Low risk. Main gap: tests duplicate the decode inline instead of exercising `acf_ajax_fetch_block()`, so they pass even if the production fix is reverted. A few stylistic nits. No security blockers.

## Non-blocking suggestions

### Tests don't actually exercise the fix
`tests/php/includes/test-blocks-query-decoding.php` — every test copies the 4-line decode snippet into the test body. The test file would still pass if someone deleted the fix from `blocks.php`. Options:
- Extract the decode into a helper (e.g. `acf_decode_block_query_arg()`) and test that.
- Or invoke `acf_ajax_fetch_block()` by priming `$_REQUEST` / nonce / caps and assert no TypeError on a string `query`.

### `$GLOBALS['wp_version']` vs `global $wp_version;`
`includes/blocks.php:917` — rest of codebase uses `global $wp_version;` (see `includes/assets.php:609`, `includes/class-acf-site-health.php:210`). Match that pattern.

### `StrictMode` PascalCase
`includes/blocks.php:917` — key name is PascalCase while neighbors (`blockTypes`, `postType`) are camelCase. This is correct because the compiled JS already reads `acf.get('StrictMode')` (PascalCase hardcoded), but worth an inline comment so a future reader doesn't "normalize" it to `strictMode` and break the feature.

### Sanitization asymmetry on `$query`
`includes/blocks.php:1035-1042` — `$query` flows through `acf_request_args()` → `acf_sanitize_request_args()` which applies `wp_kses( $v, 'acf' )` to string inputs, so the JSON string is HTML-sanitized before `json_decode` sees it. `$block` and `$context` deliberately bypass this by reading from `$_REQUEST` with a `phpcs:ignore` (lines 1025-1026). Not exploitable today because `$query` keys are only boolean-tested (`preview`, `form`, `validate`), but a latent footgun if anyone later adds a string-valued key. Consider reading `$query` the same way `$block` is read, or add a comment.

### Docblock
The inline comment in the decode block is great. The function docblock above `acf_ajax_fetch_block()` doesn't mention `$query` may arrive as a JSON string — minor.

### Test coverage gaps
No test for: JSON-encoded `true` / `null` / number (falls to `array()` — good to assert), nested object matching real `{preview,form,validate}` keys, array-passthrough via the handler path.

## Security review (20-skill framework)

**Risk: low / none.** No new trust boundary.

- **Auth/AuthZ (03):** clean — `acf_verify_ajax()` + `acf_current_user_can_edit_post` / `edit_theme_options` gate the decode (lines 997-1023).
- **Error handling (07):** clean — `json_decode` → null / non-array falls back to `array()`. Fails closed.
- **Type safety (09):** clean — `is_string()` gate + `is_array()` post-check catches JSON scalars. `wp_unslash` before `json_decode` is correct WP idiom.
- **API contracts (15):** implicit contract change — `$query` now accepts array OR JSON string. Not in docblock.
- **Web app security (16):** clean — `$query` consumed only as flag bag (`! empty( $query['preview'] )` etc.). Never echoed, never passed to `WP_Query` / `get_posts` / SQL. Verified across `includes/blocks.php`.
- **Data validation (17):** see sanitization-asymmetry note above.
- **DoS (20):** clean in practice — `json_decode` default depth 512, but attacker must already pass nonce + cap (authenticated block-editor user has higher-privileged ways to DoS). Belt-and-suspenders: add depth arg + try/catch with `JSON_THROW_ON_ERROR`. Not required.

**StrictMode info disclosure:** none — `enqueue_block_editor_assets` fires only in the authenticated block editor, which already exposes `wp_version` directly in the same localized payload (`includes/assets.php:622`). Flag only reveals WP ≥ 6.9 indirectly.

## Questions

1. Can the tests exercise `acf_ajax_fetch_block()` (or a new extracted helper)? As written they'd let a regression ship.
2. Reason for `$GLOBALS['wp_version']` over `global $wp_version;`?
3. Is `StrictMode` PascalCase deliberate to match `acf.get('StrictMode')` in the compiled JS? Worth a comment if so.

adrianmomorales:
Thanks for the thorough review @cbravobernal! Addressed in 6e03724:

**1. Tests now exercise the production code path.** Extracted the decode logic into `acf_decode_block_query_arg()` and rewrote the test file to call it directly — reverting the production fix would now break the tests. Also expanded coverage to JSON scalars (`true`/`false`/`null`/number/string), nested objects, slashed input, and non-string/non-array values (15 tests, 21 assertions).

**2. `$GLOBALS['wp_version']` → `global $wp_version;`.** No specific reason — switched to match the convention in `includes/assets.php` and `includes/class-acf-site-health.php`.

**3. `StrictMode` PascalCase is deliberate** to match `acf.get('StrictMode')` in the compiled `acf-pro-blocks.js`. Added an inline comment so a future reader doesn't normalize it.

**Other items:**
- Updated `acf_ajax_fetch_block()` docblock to mention the JSON-string case.
- Added an inline comment on the sanitization asymmetry: `$query` passes through `acf_sanitize_request_args()` (`wp_kses`) — safe today since callers only boolean-test keys, but flagged for the next person who touches it.

Let me know if you'd like me to also harden the `json_decode` call (depth arg + `JSON_THROW_ON_ERROR`) — left out as the review marked it not required, but happy to add it.

cbravobernal:
@adrianmomorales I tried to reproduce the errors by adding a v2 block and could not replicate it. Would you mind sharing a a video without / with the fix?

I'm trying to avoid AI slop 😅

cbravobernal (commented):
Thanks — this is a well-documented PR and CI is fully green. Two things need resolving before this can land, though:

**On the StrictMode half:** the premise that the flag is never set isn't quite right — `acf.data.StrictMode` is already set in `includes/assets.php:650`, gated on `SCRIPT_DEBUG && WP 6.6+` (matching Gutenberg historically enabling React StrictMode only in development). This PR adds a *second* `StrictMode` key via the block-assets localize call, which merges into the same `acf.data` and overrides the first — so on WP < 6.9 it would force the flag to `false` even with `SCRIPT_DEBUG` on, regressing the existing dev-mode handling.

If WP 6.9+ really enables StrictMode in production, the right fix is to adjust the existing gate in `includes/assets.php` (one source of truth), not to add a duplicate key. Since SCF tracks ACF upstream, it would also help to confirm how ACF Pro ≥ 6.8 handles this on WP 6.9 — we want to stay aligned with their mechanism. Could you point to a WP changeset/Gutenberg PR showing StrictMode became unconditional in 6.9?

**On the query-decoding half:** `$block` and `$context` are explicitly JSON-decoded because the JS sends them as JSON strings, but `query` has historically arrived as a regular form-serialized array. Could you link an issue or share repro steps (block type, WP/PHP versions, request payload) showing `query` arriving as a JSON string? The defensive decode is cheap, but we'd like to understand which code path serializes it that way before adding divergence from upstream in `includes/blocks.php`.

*Review run via Claude Code, reviewed by @cbravobernal.*

tapheret2 (commented):
### Review: Fix block editor rendering: query decoding and StrictMode

**Files:** `includes/blocks.php`, `tests/php/includes/test-blocks-query-decoding.php`  
**Size:** +178 / -0

**Notes**

**Nits**

Thanks @adrianmomorales — independent review pass on the diff.

PR #414, opened 2026-04-21
Map ACF dependency slug to SCF

georgestephanis:
Add scf_map_plugin_dependency_slug() to map the 'advanced-custom-fields' dependency slug to 'secure-custom-fields', allowing plugins that check for ACF to recognize SCF. The function is hooked to 'wp_plugin_dependencies_slug' and wrapped in a function_exists guard.

## Use of AI Tools

Used Copilot to draft a quick implementation of a core filter so I didn't have to pull up my own prior examples or sort out where in the codebase to land it. 🙃

### DIFF
--- a/secure-custom-fields.php
+++ b/secure-custom-fields.php
@@ -840,6 +840,24 @@ if ( ! class_exists( 'ACF' ) ) {
 	acf();
 } // class_exists check
 
+if ( ! function_exists( 'scf_map_plugin_dependency_slug' ) ) {
+	/**
+	 * Maps ACF dependency slugs so SCF satisfies plugins requiring ACF.
+	 *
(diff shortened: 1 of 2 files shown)


cbravobernal:
Pushed a quick phpunit case on top so we don't regress the mapping. Ran the full class locally, passes — phpcs/phpstan clean too.

cbravobernal:
Sorry, jumped the gun pushing directly — feel free to drop, squash, or keep the test commit however you prefer.

cbravobernal (approved):
LGTM!

georgestephanis:
Feel free to add as far as I'm concerned.  Tests always welcome! :)

PR #445, opened 2026-06-11
Remove Codecov from CI

priethor:
## What

Removes Codecov from the CI workflows. PHPUnit, JavaScript unit tests, and E2E tests still run, but they no longer spend time generating and uploading coverage reports. The Codecov config file is gone too.

## Why

We are not using Codecov for these checks anymore, so there is no reason to keep the extra coverage work in CI or during the default `wp-env` startup.

## Testing Instructions

1. Check the PHPUnit, JavaScript Unit Tests, and E2E workflow runs for this PR.
2. Confirm the tests still run.
3. Confirm there are no Codecov upload steps or Codecov PR checks.
4. Start `wp-env` and confirm it does not run the PCOV installer or load the PHP coverage collector by default.

## Use of AI Tools

I used Codex and Claude for most of the CI cleanup. I chose the approach, reviewed the diff, and tested it locally.


cbravobernal (approved):

priethor:
No Codecov report/comment on this brach 🙂

PR #466, opened 2026-06-12
Performance: 20x fewer field-lookup queries, screen-gated heavy assets, 4MB lighter zip

cbravobernal:
> **Superseded by #471 (safe bundle) + #472 (sibling-field priming).** Split per review so the lower-risk changes can merge without waiting on multilingual testing of the priming. This PR can be closed once the two successors are open for review.

Implements the highest-impact, lowest-risk items from a measured performance audit. All changes are behavior-identical: the full battery (PHPUnit / Jest / E2E / PHPStan) passes without modifying a test.

## PHP runtime
- **Prime sibling field definitions** when a field key resolves (`includes/acf-field-functions.php`): `get_field()` over a 50-field template drops from ~101 SQL queries / ~17ms to ~5 / ~2.8ms on cache-less hosts. The priming query runs with `suppress_filters => false`, matching the per-key lookup it replaces, so per-language/context query filters (WPML, Polylang, ACFML) still run — no hook is bypassed. Opt-out via `acf/prime_field_group_fields`.
- **Decode each local JSON file once per request** instead of 4× (`includes/local-json.php`): ~44ms → ~16ms at 200 files.
- **Per-request cache for `acf_get_option_meta()`** with invalidation via core option hooks. (Note: this PR keeps the `esc_like()` escaping that landed on `trunk` in #462 — the cache wraps that hardened query, it does not revert it.)
- `acf_decode_taxonomy_term()`: raw SQL → `get_term()`.
- Blocks save-path early bail when no block types are registered.
- Gate the escaped-html notice script to when the notice renders.
- Memoize the WP version check in `acf_is_using_datastore()`.

## Assets / enqueue
- `scf-bindings` moved to `enqueue_block_editor_assets` (was pulling the wp-editor stack onto classic admin + front-end `acf_form` pages).
- Command-palette scripts gated on `wp-commands` being enqueued.
- Drop `wp-polyfill` from all handles (webpack `injectPolyfill`). Compiled assets are gitignored and rebuilt at release (`bin/prepare-release.php`), so this takes effect on the next release build.
- Fix pro CSS min suffix (production was shipping unminified CSS).
- Gate `acf-pro-ui-options-page` with `is_admin()`.
- Fix `acf-dark` stylesheet 404.
- Trim release zip (exclude `assets/src` + source maps): −4.2 MB.

## Verification (on this branch, merged up to current `trunk`)
- `composer test:php`: OK (2823 tests, 21565 assertions). `npm run test:unit`: 951. `composer test:phpstan`: clean. Re-verified independently after the hook-safe priming rework.
- The `esc_like` regression test from #462 passes here (security fix preserved through the merge).
- New code is stamped `@since SCF 6.9.0` (the targeted milestone; 6.8.9 already shipped without it).

## Use of AI Tools
Initial implementation authored by Claude Code (Claude Fable 5) under human direction; reviewed and revised (hook-safe sibling priming, `@since` correction) by Claude Code (Claude Opus 4.8).


cbravobernal:
Superseded by #471 (safe perf/asset bundle) and #472 (hook-safe sibling-field priming), which split this PR by risk so the lower-risk changes can land without waiting on multilingual testing of the priming. Closing in favor of those two.

PR #475, opened 2026-06-17
Guard include arg in acf_get_users to avoid undefined array key warning

### LINKED ISSUE #457: acf_get_users() raises 'Undefined array key' warning when called without an include arg
## Description

`acf_get_users()` reads `$args['include']` without `isset()` at `includes/acf-user-functions.php:20`. Any call without an `include` argument that finds users raises an "Undefined array key" warning on PHP 8+ (notice on PHP 7.4).

## Reproduction

Repro test in #450: `tests/php/includes/functions/test-acf-user-functions.php::test_acf_get_users_warns_without_include_arg` (captures the diagnostic via error handler on both PHP 7.4 and 8+).

## Suggested fix

Guard with `! empty( $args['include'] )` / `isset()` before reading.

Found during the 2026-06 test campaign (see PR #450).

SAY-5:
Fixes #457. `acf_get_users()` reads `$args['include']` directly when reordering results, so any call without an `include` arg that finds users raises an "Undefined array key" warning on PHP 8+ (notice on 7.4). This guards the read with `! empty()` before the reorder runs.

The regression test added in #450 (`test_acf_get_users_warns_without_include_arg`) documented the warning; I flipped it to assert no warning is raised once the key is guarded.

### DIFF
--- a/includes/acf-user-functions.php
+++ b/includes/acf-user-functions.php
@@ -17,7 +17,7 @@ function acf_get_users( $args = array() ) {
 	$users = get_users( $args );
 
 	// Maintain order.
-	if ( $users && $args['include'] ) {
+	if ( $users && ! empty( $args['include'] ) ) {
 
 		// Generate order array.
 		$order = array();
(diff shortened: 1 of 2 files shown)


cbravobernal (approved):
LGTM.
Thanks for your contribution!

PR #480, opened 2026-06-22
Tests: Fix older WordPress E2E compatibility

### LINKED ISSUE #405: E2E test failures with older WP versions
In https://github.com/WordPress/secure-custom-fields/pull/404, I tentatively (and temporarily) enabled e2e tests for all WP versions starting with 6.2, which is the minimum version required per the plugin header: https://github.com/WordPress/secure-custom-fields/blob/183d93e21f032c60afcacbfcc0448bc9f6551525/secure-custom-fields.php#L15

I expected Command Palette tests to fail with 6.2, since the Command Palette was only introduced in 6.3. (I'm addressing that in https://github.com/WordPress/secure-custom-fields/pull/404.)

However, Command Palette tests also failed in 6.3, alongside a lot of Field Type tests (`tests/e2e/field-type-*.spec.ts`).
One error I've seen in the browser console is

```
Uncaught TypeError: Cannot read properties of undefined (reading 'jsx')
```

which is due to a missing `ReactJSXRuntime` dependency. The latter was [introduced in WP 6.6](https://make.wordpress.org/core/2024/06/06/jsx-in-wordpress-6-6/); we'll likely need to [polyfill it](https://github.com/WordPress/gutenberg/issues/62202#issuecomment-2156796649) in order to support earlier versions.

(Alternatively, we could bump the minimum required version to WP 6.8.)

[Other test failures:](https://github.com/WordPress/secure-custom-fields/actions/runs/24191830747?pr=404)

- 6.3, 6.4, 6.5:
  - `tests/e2e/block-testimonial.spec.ts:31:2 › SCF Block > Testimonial › should use testimonial block with fields`
  - `tests/e2e/block-v3-features.spec.ts:161:2 › SCF Block V3 Features › should validate required fields on publish`
  - `tests/e2e/command-palette.spec.ts`: Multiple errors
- 6.5, 6.6:
  - `tests/e2e/block-testimonial.spec.ts:31:2 › SCF Block > Testimonial › should use testimonial block with fields`
  - `tests/e2e/block-v3-features.spec.ts:36:2 › SCF Block V3 Features`
    - `should render v3 block in preview-only mode`
    - `should update preview when sidebar fields change`
    - `should save v3 block data correctly`
- 6.5, 6.6, 6.7.:
  - `tests/e2e/block-bindings-site-editor.spec.ts:24:2 › Block Bindings in Site Editor › should bind text field to paragraph block in site editor`

and more in 6.3 and 6.4.

priethor:
## What

Closes #405.

This fixes the older WordPress editor failures that showed up while testing SCF against intermediate core versions. Command palette bundles now load the dependencies generated at build time and get a JSX runtime fallback when WordPress does not provide one. Block V3 fields remain editable in older editor UIs, and the Site Editor bindings coverage now only runs where SCF's bindings UI can actually work.

## Why

SCF still supports older WordPress versions, but several editor paths were relying on JavaScript APIs or UI details that changed between WordPress 6.3 and 6.8. The fixes keep those flows working without permanently expanding the GitHub Actions matrix.

## How

- Loading generated script asset metadata before registering SCF scripts.
- Providing a React JSX runtime fallback for older WordPress versions.
- Keeping the expanded block editor usable when sidebar fields are not available.
- Avoiding the legacy bindings editor script before WordPress 6.7, where the required JS APIs are unavailable.

## Testing Instructions

1. In a WordPress 6.3 to 6.6 environment, open the post editor and confirm SCF command palette commands appear.
2. Add a Block V3 test block, edit its fields, close the expanded editor, save, and reload the post.
3. In WordPress 6.7 or newer, open the Site Editor, select a bindable paragraph block, and confirm the SCF field binding control is available.
4. In WordPress 6.6 or older, open an editor screen that loads SCF input assets and confirm the page does not show a block bindings JavaScript error.

## Use of AI Tools

I used Codex to help apply the fixes in this PR, and I reviewed the changes.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -28,7 +28,7 @@ jobs:
             matrix:
                 event: ['${{ github.event_name }}']
                 shard: [1, 2, 3, 4]
-                wp: ['6.2', '6.3', '6.4', '6.5', '6.6', '6.7', '6.8', 'latest', 'trunk']
+                wp: ['6.2', '6.8', 'latest', 'trunk']
                 exclude:
                     # On PRs: Skip minimum supported WP version for faster feedback
                     - event: 'pull_request'
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -28,7 +28,7 @@ jobs:
             matrix:
                 event: ['${{ github.event_name }}']
                 shard: [1, 2, 3, 4]
-                wp: ['6.2', '6.8', 'latest', 'trunk']
+                wp: ['6.2', '6.3', '6.4', '6.5', '6.6', '6.7', '6.8', 'latest', 'trunk']
                 exclude:
                     # On PRs: Skip minimum supported WP version for faster feedback
                     - event: 'pull_request'
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -1,5 +1,6 @@
 {
   "$schema": "https://schemas.wp.org/trunk/wp-env.json",
+  "phpVersion": "8.2",
   "plugins": ["."],
   "env": {
     "tests": {
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -1,6 +1,5 @@
 {
   "$schema": "https://schemas.wp.org/trunk/wp-env.json",
-  "phpVersion": "8.2",
   "plugins": ["."],
   "env": {
     "tests": {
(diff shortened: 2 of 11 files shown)


priethor:
Confirming e2e tests fail in 6.3-6.7
image

priethor:
e2e test snow pass in all versions:
image

Reverting the actions to the previous matrix now.

PR #485, opened 2026-06-24
Docs: Update WP requirements to 6.2

mkismy:
The plugin currently requires WordPress 6.2 or later but the document says 6.0 or later.

I updated the document to match the current requirements.

## Use of AI Tools

None


cbravobernal (approved):
LGTM

PR #495, opened 2026-07-09
Copy: make strings reported in #479 translatable

### LINKED ISSUE #479: A string is not translatable
While I was translating strings alongside the plugin UI, I noticed that a text string "Allow content editors to select multiple values" does not appear in translate.wordpress.org.

This is where I spotted on the UI.

- Add new field group
- Add new field
- Select field type "Select" or "User"
- Scroll down to the last option (just above "Close field")

Image

kmfoysal06:
Fixes #479 by wrapping the untranslated text string in the `__()` translation helper.

Closes #479

### DIFF
--- a/includes/fields/class-acf-field-page_link.php
+++ b/includes/fields/class-acf-field-page_link.php
@@ -483,7 +483,7 @@ if ( ! class_exists( 'acf_field_page_link' ) ) :
 			acf_render_field_setting(
 				$field,
 				array(
-					'label'        => __( 'Allow Archives URLs', 'secure-custom-fields' ),
+					'label'        => __( 'Allow Archive URLs', 'secure-custom-fields' ),
 					'instructions' => '',
 					'name'         => 'allow_archives',
 					'type'         => 'true_false',
@@ -495,7 +495,7 @@ if ( ! class_exists( 'acf_field_page_link' ) ) :
 				$field,
 				array(
 					'label'        => __( 'Select Multiple', 'secure-custom-fields' ),
-					'instructions' => 'Allow content editors to select multiple values',
+					'instructions' => __( 'Allow content editors to select multiple values', 'secure-custom-fields' ),
 					'name'         => 'multiple',
 					'type'         => 'true_false',
 					'ui'           => 1,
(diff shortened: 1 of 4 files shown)


tapheret2 (commented):
### Review notes on “hotfix: translation issue #479”

Touched files (1): `includes/fields/class-acf-field-select.php`
Diff size (approx): +1 / -1

- Change set is focused. I don’t see an obvious correctness red flag from the diff alone.

**Questions / nits**
- Does this need a changelog / release note entry for the target version?
- Any user-facing behavior change that should be called out in the PR body?

Thanks @kmfoysal06 for the contribution — left this as an independent review pass.

kmfoysal06:
Thanks @tapheret2  for the review! This doesn't introduce any functional behavior changes. The only user-facing difference is that the previously hardcoded string is now translatable and will appear correctly in localized installs.

priethor:
Hi @tapheret2,

Thanks for the interest in SCF; it's always good to see more eyes on the PRs.

One thing we’ve noticed, though: the same layout and the same general questions come up across quite a few unrelated PRs, which tends to add more noise than signal. A review that isn’t tied to the specific diff doesn’t give us much to act on. Here, for example, the PR description already explains the user-facing effect (it makes a hardcoded string translatable), so the questions were pretty much answered already.

If something concrete in a diff looks off, we’d genuinely like to hear it. Otherwise, there’s really no need to leave a review just to leave one. Thanks for understanding.

priethor (changes requested):
Thanks @kmfoysal06! Looks good.

One catch before it goes in: the string is hardcoded in a few other field types too, and #479 calls out the User field specifically, so this on its own won’t close it. Could you grab these while you’re at it?

- class-acf-field-user.php
- class-acf-field-post_object.php
- class-acf-field-page_link.php

Same fix as here, and we’re good to go. 🙂

kmfoysal06:
@priethor Thanks for the review! Sure, I’ll update those field types as well and push the changes shortly.

kmfoysal06:
@priethor  Hello, I just pushed the updates for the remaining files you mentioned. I also confirmed that those files don't have any other missing translatable strings.

priethor (approved):
Thanks, looks good to me. The `Archive` rename is slightly out of scope, but it's a good fix nevertheless.

kmfoysal06:
@priethor Thanks for taking the time to review the PR and for the approval. I really appreciate it!

priethor:
Thank you for contributing!

PR #500, opened 2026-07-14
#499 UI: Align URL Field Globe Icon Inside the URL Input Box.

kmfoysal06:
This PR is to resolve #499 . the issue is the globe icon in url field does not aligned in center. I added css to make it center aligned.
``` css 
   .acf-url i {
      top: 50%;
      transform: translateY(-50%);
} 
``` 

Closes #499

PR #504, opened 2026-07-15
Backport: ACF 6.8.6 changes

ockham:
## Description

Ports the ACF 6.8.6 point release (2026-07-14) into SCF. Upstream range: ACF `6.8.5 → 6.8.6` (ACF 6.8.5 was verified to be fully covered already by earlier SCF work: #462, #467, #491 and related commits).

All seven changelog fixes from upstream are included:

- **Location rules**: guard against missing rule `value`/`operator` keys to avoid PHP warnings on page load (`includes/locations/*`). Ported from the free ACF repo (canonical for non-Pro files).
- **WP 7.0+ field appearance**: add `acf-admin-7-0` body class and port the improved URL/Number/Select field CSS. Rebased after #506: the styles are now upstream's own 6.8.6 `_fields.scss` patch applied to the partial (its `.acf-url` hunk was already on trunk via #500/#506), leaving `_fields.scss` byte-identical to upstream 6.8.6.
- **Blocks auto inline editing**: only substitute placeholder values for fields fetched against the block's own post id, and save/restore the render context so nested block renders don't clobber the parent (`includes/blocks-auto-inline-editing.php`; the stale, unloaded `pro/` copy is untouched).
- **WooCommerce HPOS compatibility mode**: register `acf/form-post/skip_save` so `save_order()` stays the single save entry-point and the backfill-triggered `save_post` no longer consumes the ACF nonce. Adapted to cover all registered order types via `wc_get_order_types()`, matching SCF's existing multi-order-type support.
- **Validation JS**: scope submit-click validation to the button's own form so unrelated UI (e.g. the Classic Editor link inserter) no longer triggers premature validation of a Link field; use `validity.valid` instead of `checkValidity()` to avoid dispatching native `invalid` events.
- **Blocks V2/V3**: don't crash on attribute values that merely start with `[` or `{` (e.g. oEmbed titles) — `JSON.parse` is now try/caught in both the V2 parser (`_acf-blocks.js`) and V3 (`jsx-parser.js`). Two jsx-parser tests asserting the old throwing behavior were updated.
- **Blocks V3**: deserialize Google Map field values after `acf.serialize()` so they aren't double-encoded in block attributes (`block-edit.js`).
- **DOMPurify**: bump 3.3.0 → 3.4.11 to match upstream's bundled upgrade.

The two block fixes only ship minified upstream; both bundle versions were pretty-printed and diffed (a string-literal set comparison confirmed these were the only substantive changes), then implemented in maintainable source — no decompiled output was pasted.

**Intentionally not ported**: the ACF email opt-in marketing banner (admin page, view, JS, CSS, images — collects emails for ACF's newsletter, not applicable to SCF), readme/version bumps, and `lang/**` translation churn (SCF has its own translation workflow).

## Testing

- `php -l` clean on all changed PHP files; `npm run build` passes (DOMPurify 3.4.11 bundles cleanly)
- phpcs-changed pre-commit hook passes; no new PHPStan or ESLint findings vs trunk (ESLint count actually drops by 3)
- PHPUnit: 2845 tests, 21611 assertions, all passing
- Jest: 951 tests, all passing
- E2E (local wp-env): 249/253 passed; the 4 failures (`post-type-admin.spec.ts` ×3, `block-bindings-site-editor.spec.ts` ×1) reproduce identically on trunk and are unrelated to this PR

## Use of AI Tools

This PR was created with assistance from Claude Code (Anthropic).

### DIFF
--- a/assets/src/js/_acf-validation.js
+++ b/assets/src/js/_acf-validation.js
@@ -746,13 +746,20 @@
 	 *
 	 * @date	20/10/2021
 	 * @since	ACF 5.11.0
+	 *
+	 * @param {jQuery} [$form] - Optional form element to scope the input search to. If omitted, searches the entire document.
 	 */
-	var ensureInvalidFieldVisibility = function () {
-		// Load each ACF input field and check it's browser validation state.
-		var $inputs = $( '.acf-field input' );
+	const ensureInvalidFieldVisibility = function ( $form ) {
+		// Load each ACF input field and check its browser validation state.
+		// Scope to the given form if provided, otherwise search the entire document.
+		const $inputs = $form
+			? $form.find( '.acf-field input' )
+			: $( '.acf-field input' );
 		$inputs.each( function () {
-			if ( ! this.checkValidity() ) {
-				// Field is invalid, so we need to make sure it's metabox is visible.
+			// Use validity.valid (a property) instead of checkValidity() (a method) to avoid
+			// dispatching the native 'invalid' event, which can trigger ACF validation prematurely.
+			if ( ! this.validity.valid ) {
+				// Field is invalid, so we need to make sure its metabox is visible.
 				ensureFieldPostBoxIsVisible( $( this ) );
 			}
 		} );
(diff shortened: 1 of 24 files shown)


ockham:
Rebased onto trunk after #506 merged. What changed relative to the previously reviewed state:

- The WP 7.0 field-appearance CSS moved from the (now gone) flattened `acf-input.scss` into `_fields.scss`, as upstream's own 6.8.6 patch — the file is now byte-identical to upstream 6.8.6, since trunk already carried the `.acf-url` hunk (#500/#506).
- The follow-up commit `abe0948` (mixin rewrite of the flat CSS) is dropped as superseded; the branch is back to a single commit.
- Everything outside `assets/src/sass/` is byte-identical to the reviewed diff (`git diff <old-head> HEAD -- . ':(exclude)assets/src/sass'` shows only trunk's own #505 revert, inherited by rebasing).

PR #506, opened 2026-07-16
Restore upstream sass structure: entries import partials again

ockham:
## Description

SCF's four main sass entry files (`acf-input.scss`, `acf-global.scss`, `acf-field-group.scss`, `acf-dark.scss`) had been flattened into compiled CSS with no imports, orphaning the `_*.scss` partials. The partials kept receiving upstream syncs (e.g. the 6.8.1 backport, #428) while the flattened entries were edited separately (e.g. #500), so the two sources silently diverged — and several upstream style updates that the partials already contained never actually shipped.

This restores upstream's structure so future CSS ports are plain patches against the partials. Three commits, meant to be reviewed separately:

1. **Restore upstream sass structure** — entries and partials synced verbatim from upstream ACF **6.8.5** (trunk's current parity level), including the previously missing `_ai-admin.scss`. Upstream has used the `@use` module system for these files since before 6.8.5; we now match it exactly.
2. **Strip commercial upsell CSS** — SCF ships no licensing/upgrade UI, so upstream's commercial blocks are removed from the synced partials, each marked with an `// SCF:` comment for future ports: the `pro-upgrade` module, `#acf-license-information` panel, `.acf-pro-inactive-license` row states, WP Engine toolbar branding, and license page icons. Matches what SCF's shipped CSS already excluded.
3. **Preserve #500** — the URL globe-icon centering re-applied to `_fields.scss`. Upstream adopts the identical change in ACF 6.8.6, so this deviation disappears at the next sync.

## Shipped CSS changes

The built output is intentionally *not* byte-identical to trunk — the restructure fixes staleness bugs where shipped CSS had missed upstream updates (deltas vs trunk: `acf-dark` 0 lines, `acf-input` 34, `acf-field-group` 238, `acf-global` 670, all enumerated below):

- Schema/GEO select2 dropdowns (`schema-type/property/output-format-select-results`) are now styled — they shipped completely unstyled despite the GEO feature being ported
- Google Map toolbar buttons get their intended styling (missing since upstream 6.8.1)
- Copyable field-key hover/copied affordances in list tables
- `ai_description` field-group option gets the same width constraint as its siblings
- Block-editor sidebar field notices use upstream's corrected margins
- Field-object message settings and `acf_post_type_supports` selectors modernized to upstream's current form
- Removed dead CSS for the old "Upgrade to Pro" gradient button (`.acf-admin-toolbar-upgrade-btn`) — no markup references it
- The `#acf-admin-tools h1` hide rule keeps upstream's `:not(...)` exception clause (functionally identical in SCF, which never renders those classes)

Verified: built CSS contains zero `license-information` / `pro-inactive-license` / `wp-engine` / `upgrade-btn` selectors.

## Notes

- The `pro/` sass entries are untouched: upstream's PRO repo ships no sass sources, so there is nothing to align them to.
- Two pre-existing sass deprecation warnings (`darken()` in upstream's own `_global.scss`) are left as-is to keep the partial verbatim.
- This conflicts with #504's CSS commit; if this merges first, that commit can be replaced by a plain `git apply` of upstream's 6.8.6 `_fields.scss` patch — exactly the workflow this PR enables.


assets/src/sass/_field-picker.scss
```diff
@@ -141,6 +144,30 @@
...
+					background-image: url("../../images/pro-chip.svg");
+					background-repeat: no-repeat;
+					height: 24px;
+					width: 28px;
+
+					&.not-pro {
+						background-image: url("../../images/pro-chip-locked.svg");
+						width: 43px;
+					}
+				}
```
ockham: This looks like it's related to some sort of "Upgrade to Pro" element. We probably don't need it.

ockham:
Two follow-up commits pushed after review of the remaining "pro" selectors in the built output:

- **341e882** strips PRO-tier CSS the first upsell pass missed, none of which any SCF markup or script can render: the `.field-type-requires-pro` chip (its `pro-chip*.svg` images don't even exist in SCF), `.field-type-upgrade-to-unlock` / `.acf-btn-pro` (the JS branch showing them is permanently disabled), `.acf-pro-label(-field-type)`, `.acf-options-pages-preview-upgrade-button`, `.acf-ui-options-page-pro-features-actions`, `.acf-requires-pro`, and the `.acf-logo.pro` variant. `#acf-upgrade-notice` stays — despite the name it's the database-upgrade notice.
- **44a4860** deletes the now-orphaned `_pro-upgrade.scss` (492 lines, zero bytes in built CSS); the `acf-global.scss` comment marks the module as intentionally absent so future syncs don't reintroduce it.

The only "pro" references left in built CSS are the inert `:not(.acf-field-group-pro-features-title…)` exception clauses kept for selector parity with ACF.

assets/src/sass/_field-picker.scss
ockham: Adressed by https://github.com/WordPress/secure-custom-fields/pull/506/commits/341e882e0835d0e147f4fe117fce05dc79be1664.

ockham:
Reviewer note from visual inspection: two of the shipped-CSS delta families in this PR are **deliberately forward-looking** — their rules match upstream but nothing in SCF applies the classes yet, so they produce no visual change:

- **Schema/GEO picker dropdowns** (`schema-*-select-results` in `_admin-inputs.scss`): upstream tags the dropdowns from `_acf-internal-post-type.js`, whose schema code SCF hasn't ported. Follow-up: #507
- **List-table copyable field keys** (`.column-acf-key .copyable` in `_list-table.scss`): the markup + click handling come from upstream's `_acf-copyable.js`, which SCF lacks entirely. The field-group *editor* copy affordance is unrelated and already worked before this PR. Follow-up: #508

Keeping the CSS in this sync is intentional — it keeps the partials verbatim-upstream, and each follow-up JS port then lights the feature up without touching CSS. The families with visible changes (`ai_description` width, Google Map toolbar buttons, dead upgrade-button removal, etc.) are unaffected by this note.

PR #537, opened 2026-08-19
Count PHP-registered field groups in Site Health

### LINKED ISSUE #456: Site Health never counts PHP-registered field groups ('PHP' vs 'php' comparison); legacy site-health class is dead code
## Description

1. **PHP-registered field group count is always 0.** The Site Health data counts PHP field groups with `'PHP' === $field_group['local']`, but `acf_add_local_field_group()` registers groups with `'local' => 'php'` (lowercase). The comparison never matches. The live class has this bug at `src/Site_Health/Site_Health.php:520`; the same comparison exists in the legacy `includes/class-acf-site-health.php:340`.
2. **`includes/class-acf-site-health.php` appears to be dead code.** The plugin only instantiates `SCF\Site_Health\Site_Health` from `src/`. The legacy class additionally treats `acf_get_post_types()` / `acf_get_taxonomies()` (which return name strings) as arrays (`$post_type['local']` at `:253-254`), making its counts meaningless — further evidence it is unused and could be removed rather than fixed.

## Reproduction

Repro test in #450: `tests/php/includes/test-class-acf-site-health.php` (`php_field_groups` asserted `'0'` with a PHP-registered group present, carrying the NOTE comment).

## Suggested fix

Case-insensitive compare (or compare against `'php'`) in `src/Site_Health/Site_Health.php`; remove the legacy `includes/class-acf-site-health.php` after confirming nothing external instantiates it.

Found during the 2026-06 test campaign (see PR #450).

faisalahammad:
## Summary

`acf_add_local_field_group()` registers field groups with `local => 'php'` (lowercase), but the Site Health debug info counted PHP-registered groups by matching the uppercase value `'PHP'`. The comparison never matched, so "Registered Field Groups (PHP)" always reported 0.

Changes in this PR:
- Fix the comparison in `src/Site_Health/Site_Health.php` to match the lowercase `'php'` value, consistent with the JSON count which already uses `'json'`.
- Remove the dead legacy `ACF_Site_Health` class (`includes/class-acf-site-health.php`). It was never loaded by the plugin bootstrap and also carried the same bug plus unrelated latent bugs.
- Replace its legacy-only test with `tests/php/src/test-site-health.php`, which covers the production class and locks in the corrected PHP field group count.

## Test plan

- `composer test:php` passes, including the new regression test that asserts PHP-registered groups are counted (not `'0'`).
- PHPStan and PHPCS are clean on the changed files.
- Manual check: register a field group via `acf_add_local_field_group()` in code, open Tools > Site Health > Info > Secure Custom Fields, and confirm "Registered Field Groups (PHP)" shows the expected non-zero count.

Closes #456

## Use of AI Tools

This pull request was created with the assistance of AI coding tooling (Claude Code). All changes have been reviewed and I take responsibility for what the tooling produced.


faisalahammad:
Please rerun the failed jobs for PR #537. The three E2E shards timed out while installing Playwright browsers, before any tests ran. The rerun command requires repository-admin permission.

Issue #156, opened 2025-05-20
Migrate the SCF post editing pages to leverage Core Components


priethor:
SCF currently uses custom PHP form and field helper functions to build interfaces for creating and editing Field Groups, Post Types, Taxonomies, and Options Pages post types. This project aims to rebuild the SCF's post types editing logic (`post.php`) by leveraging native WordPress components such as DataForms and other core UI elements. The goal is to modernize and simplify these editing pages, improving maintainability and user experience by aligning with WordPress core design patterns.

### Additional Context
  - It should be started after #142 begins.  
  - It should be implemented with reusability in mind, e.g., to integrate them in the block editor for quick editing.
  - We should start with Taxonomies, as it is closely related to #155 and simpler than other types.
  - It should be introduced as an opt-in Beta feature.

### Tasks

- [ ] Migrate the Taxonomies editing page to native components
- [ ] Migrate the Field Groups editing page to native components
- [ ] Migrate the Post Types editing page to native components
- [ ] Migrate the Options Pages to native components
- [ ] Explore embedding these in the Editor, for example to Edit Taxonomies as in #155

Issue #142, opened 2025-05-20
Migrate Admin Listing Pages (edit.php) to DataViews


priethor:
The plugin currently provides users with several index pages: Field Groups, Post Types, Taxonomies, Options Pages, and listings for every Custom Post Type created through the plugin. To modernize these experiences and align with WordPress core’s direction, all of the plugin’s listing pages should be migrated to DataViews. This will provide a more user-friendly and visually appealing interface, richer out-of-the-box features, and reduce future maintenance burden.

### Additional Context

- **Dependencies:** Heavy dependence on the dataviews component. Since `@wordpress/dataviews` is not actively worked on, it should be based on the more active [`@automattic/dataviews` fork](https://www.npmjs.com/package/@automattic/dataviews).
- **Considerations:**
  - These pages will initially be opt-in as a Beta Feature.
  - Check the implications this change has on plugins that extend SCF.
- **Future iterations:** Implement quick-edit

PR #546, opened 2026-09-22
feat(taxonomy): add opt-in DataForm editor

faisalahammad:
Adds an opt-in beta feature that renders the taxonomy editor Basic Settings with DataForm instead of the classic PHP fields. This is the first slice of #156 and covers taxonomies only, as the issue suggests. Field groups, post types and options pages are follow-ups.

The classic Basic Settings markup stays in the page and is hidden only after the editor mounts. That keeps the Post Types selector, the Advanced Configuration toggle, the taxonomy key slugification and any `acf/taxonomy/basic_settings` additions working, and it means a script failure leaves the original fields usable. Saving is untouched: the existing nonce, capability check, AJAX key validation and `acf_update_internal_post_type()` are still the only write path.

The editor uses `@wordpress/dataviews` and is limited to WordPress 7.0 and above, where the package is available. Note that #142 mentioned `@automattic/dataviews` back in May 2025, while the current WordPress docs use `@wordpress/dataviews`. Happy to switch if the listing work in #142 lands on the other package.

How to test:

1. Go to Field Groups > Beta Features and enable Taxonomy Editor.
2. On WordPress 7.0 or above, edit a taxonomy. The Plural Label, Singular Label, Taxonomy Key, Public and Hierarchical fields are now rendered by DataForm.
3. Change those fields and save. The values persist on reload.
4. Turn a toggle off and save, for example make a public taxonomy private. It stays off after reload.
5. Confirm Post Types and Advanced Configuration still work.
6. Disable the beta feature, or run WordPress below 7.0. The classic form shows again.

Two things worth knowing before review: the taxonomy key slugifies on blur of the classic singular label field, which is now hidden, so editing only the DataForm label will not auto-fill the key. The key is still editable in DataForm and saves as typed. Built assets are gitignored, so a source checkout needs `npm run build` before the editor loads.


faisalahammad:
## CI Fix Summary

Updated `tests/e2e/fixtures.js` to support the WordPress trunk preview menu label change.

- Handles both `View` and legacy `Preview` editor buttons.
- Matches both `Preview in new tab` and `Preview (opens in a new tab)` menu labels.
- Fixes the WP trunk `openPreviewPage()` timeout affecting shards 2/4, 3/4, and 4/4.

Validation: lint, Prettier, and diff checks passed. CodeRabbit was unavailable because the review limit was exhausted and no organization seat is assigned.

Commit: `0f7a15a`

PR #544, opened 2026-09-22
fix(export): register post meta in PHP export

faisalahammad:
Generated PHP now includes `register_post_meta()` calls for fields that are eligible, so exported field groups work with block bindings in block themes without manual registration. Previously the export only emitted `acf_add_local_field_group()` and `register_post_type()`, so block bindings reported the field as not registered and it did not show in the bindings picker.

Closes #384

### What gets registered

A field is exported as post meta only when all of these are true:

- The field group has Show in REST API enabled.
- The field group targets a concrete post type, meaning every location group is a single `post_type == <slug>` rule. Groups with mixed or broader rules are skipped rather than guessed.
- The field is not disabled for bindings, and its field type supports bindings.
- The field type is REST enabled.
- The field resolves to exactly one scalar REST type (`string`, `boolean`, `integer`, `number`). Multi value fields such as checkbox, repeater, gallery, or relationship are skipped, since they cannot be registered as single native meta safely.
- The field name is not a protected meta key (no leading underscore).

The output is added to the existing export as a separate block:

```php
add_action( 'init', function() {
	register_post_meta( 'book', 'book_title', array(
		'single'       => true,
		'type'         => 'string',
		'show_in_rest' => true,
	) );
} );
```

The existing `acf_add_local_field_group()` output is unchanged, so this is additive.

### Files changed

- `includes/acf-field-group-functions.php` - helper functions to resolve concrete post types from location rules, decide which fields can be registered, and build the exported PHP.
- `includes/admin/tools/class-acf-admin-tool-export.php` - prints the post meta block alongside the field group export.
- `tests/php/includes/test-acf-field-group-functions.php` - tests for scalar export, multiple post types, skipped fields, and ambiguous locations.

### How to test

1. Create a field group on a custom post type, for example a `book` post type.
2. Enable Show in REST API for the group and add a text field.
3. Use Tools > Export to generate PHP for that group.
4. Confirm the output contains a `register_post_meta( 'book', ... )` call for the field.
5. Load the exported PHP and open a block theme template. The field should appear in the Block Bindings attributes list.

### Testing done

- `composer test:php` passes: 2947 tests, 22753 assertions.
- `composer lint:php` and the phpcs changed lines check pass on the changed files.
- Full build runs clean.
- Manually verified the exported PHP loads and the field appears in the Block Bindings picker.

## Use of AI Tools

[...]

PR #547, opened 2026-09-22
feat(release): resolve @since version placeholders at release time

### LINKED ISSUE #151: Release management: automatically tag `@since` annotations with the correct version
As of SCF 6.4, every time we add a new APi, we need to manually add the specific version `@since` annotation, which can lead to inaccuracies: if there is a new release before the branch with the new API is merged, its `@since` annotations need to be updated.

SCF should have a placeholder like `{NEXT_MAJOR_VERSION}` that gets replaced with the appropriate version during the release process jobs, so that contributors just need to type `@since {NEXT_MAJOR_VERSION}`.

faisalahammad:
Closes #151

New SCF APIs can now be written as `@since SCF {NEXT_MAJOR_VERSION}` instead of guessing the version number. `composer prepare-release` replaces the placeholder with the real release version after the maintainer enters and confirms it, so annotations stay correct when a release slips and the version changes.

`Since_Resolver` in `bin/class-scf-since-resolver.php` scans `includes`, `src` and `assets/src` for PHP, JS, JSX, TS and TSX files. It only rewrites the token inside an `@since` annotation and normalises the result to `@since SCF <version>`, so legacy `@since ACF ...` tags, versions that are already resolved, and an unrelated placeholder in a string literal are left untouched. The resolver returns the list of changed files and fails loudly if a write does not go through.

The release script calls it right after the changelog confirmation, then regenerates the docs and commits only the resolved source files plus `docs/code-reference` and `docs/bin/manifest.json`. The commit is skipped when nothing changed.

Contributor docs now describe the convention in `docs/contributing/index.md`, and `docs/contributing/releases.md` documents the release step.

Two limits worth knowing: `pro/` is not scanned because it has no `@since` tags today, and the docs generator only parses PHP under `includes`, so a placeholder in `src/` or in JS is rewritten in source but does not show up in the generated code reference.

## Testing instructions

* `composer test:php` (the new `Test_SCF_Since_Resolver` covers placeholder replacement, existing ACF and SCF tags staying put, a bare placeholder outside an annotation being ignored, unsupported file types, and the no-op case)
* `vendor/bin/phpunit --filter Test_SCF_Since_Resolver tests/php`
* `composer lint:php`
* `npm run test:unit`

Manual check: copy the plugin to a temp directory, add `@since SCF {NEXT_MAJOR_VERSION}` to a PHP file under `includes` and a JS file under `assets/src`, then run the resolver against that tree. The new annotations become `@since SCF <version>`, and existing `@since ACF 6.1` and `@since SCF 6.9.5` tags are unchanged.

`bin/` and `tests/` are not part of the release zip, so this change has no effect on the distributed plugin. E2E tests were not run because nothing in the shipped runtime path changed.


faisalahammad:
## CI Fix Summary

The failing E2E checks were not caused by this PR. WordPress trunk renamed the editor preview menu item from "Preview in new tab" to "Preview (opens in a new tab)". The pinned test helper `@wordpress/e2e-test-utils-playwright` (1.32.0) looks for the old exact label, so `editor.openPreviewPage()` timed out.

This is a trunk wide breakage that hits every branch, and it is already fixed on the `fix/384-export-post-meta` branch (PR #544).

### Change

- `tests/e2e/fixtures.js`: the local `editor.openPreviewPage` override stopped calling the upstream helper when the "View" button is visible. It now clicks the View button (or Preview on WP 6.2) itself and matches the menu item with `getByRole('menuitem', { name: /Preview.*new tab/i })`, so both the old and the new label match.

Test only change. No plugin runtime code touched.

### Errors fixed

`TimeoutError: browserContext.waitForEvent: Timeout 10000ms exceeded while waiting for event "page"` at `Editor.openPreviewPage` (preview.ts:27), 39 test cases across WP trunk shards 2/4, 3/4 and 4/4.

### Validation

- `npx wp-scripts lint-js` and Prettier clean
- `npm run test:unit` pass
- `composer test:php` 2948 tests, 22755 assertions, OK
- Targeted PHPStan and PHPCS on changed files pass
- No E2E run locally. CI will confirm the trunk shards on this push.

Refs #547

